PR #10, opened 2016-09-20
interface doc: Added Security Considerations section.

mgiuca:
This summarizes the security issues that we have identified in discussions with the Chrome security team.

### DIFF
--- a/docs/interface.md
+++ b/docs/interface.md
@@ -22,11 +22,11 @@ workers) is the main method of the interface:
 
 ```WebIDL
 partial interface Navigator {
-  Promise<void> share(ShareData data);
+  [SecureContext] Promise<void> share(ShareData data);
 };
 
 partial interface WorkerNavigator {
-  Promise<void> share(ShareData data);
+  [SecureContext] Promise<void> share(ShareData data);
 };
 
 dictionary ShareData {
@@ -134,9 +134,8 @@ platform).
   asking them to select a target application (even if there is only one possible
   target). This surface serves as a security confirmation, ensuring that
   websites cannot silently send data to native applications.
-* Due to the capabilities of the API surface, implementations MAY choose to
-  restrict `navigator.share` to secure browsing contexts (such as `https://`
-  schemes).
+* Due to the capabilities of the API surface, `navigator.share` is restricted to
+  secure browsing contexts (such as `https://` schemes).
 * Use of `navigator.share` from a [private browsing
   mode](https://en.wikipedia.org/wiki/Privacy_mode) may leak private data to a
   third-party application that does not respect the user's privacy setting.
--- a/docs/interface.md
+++ b/docs/interface.md
@@ -116,3 +116,34 @@ when a share is sent to a native application, the user agent may create an
 object with `ACTION_SEND`, setting the `EXTRA_SUBJECT` to `title`. Since Android
 intents do not have a URL field, `EXTRA_TEXT` would be set to `text` and `url`
 concatenated together.
+
+## Security considerations
+
+Implementations should observe the following security and privacy advice.
+
+Web Share enables data to be sent from websites to native applications. While
+this ability is not unique to Web Share, it does come with a number of potential
+security issues that may vary in severity (depending on the underlying
+platform).
+
+* Implementations MUST NOT allow the website to learn which apps are
+  installed, or which app was chosen from `navigator.share`. This information
+  could be used for fingerprinting, as well as leaking details about the user's
+  device.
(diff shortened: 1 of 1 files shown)


docs/interface.md
```diff
@@ -116,3 +116,34 @@ when a share is sent to a native application, the user agent may create an
...
 intents do not have a URL field, `EXTRA_TEXT` would be set to `text` and `url`
 concatenated together.
+
+## Security considerations
+
+Implementations should observe the following security and privacy advice.
+
+Web Share enables data to be sent from websites to native applications. While
+this ability is not unique to Web Share, it does come with a number of potential
+security issues that may vary in severity (depending on the underlying
```
marcoscaceres: s/may/can

docs/interface.md
```diff
@@ -116,3 +116,34 @@ when a share is sent to a native application, the user agent may create an
...
+platform).
+
+* Implementations MUST NOT allow the website to learn which apps are
+  installed, or which app was chosen from `navigator.share`. This information
```
marcoscaceres: s/which apps/which native applications

Also, maybe change "Implementations" to use agent, so it's more clear that in which direction the information is moving and who is moving it... saying "Implementations" is a bit vague, IMO.

docs/interface.md
```diff
@@ -116,3 +116,34 @@ when a share is sent to a native application, the user agent may create an
...
+  could be used for fingerprinting, as well as leaking details about the user's
+  device.
+* On every call to `navigator.share`, the user MUST be presented with a dialog
+  asking them to select a target application (even if there is only one possible
+  target). This surface serves as a security confirmation, ensuring that
+  websites cannot silently send data to native applications.
+* Due to the capabilities of the API surface, websites MAY choose to restrict
```
marcoscaceres: Probably make this a must.

docs/interface.md
```diff
@@ -116,3 +116,34 @@ when a share is sent to a native application, the user agent may create an
...
+  `navigator.share` to secure browsing contexts (such as `https://` schemes).
```
marcoscaceres: Probably just link to the spec that defines secure context and don't say "(such as...)".

docs/interface.md
```diff
@@ -116,3 +116,34 @@ when a share is sent to a native application, the user agent may create an
...
+* Use of `navigator.share` from a [private browsing
+  mode](https://en.wikipedia.org/wiki/Privacy_mode) may leak private data to a
```
marcoscaceres: s/may/can (it's a statement of fact, not a conformance requirement)

docs/interface.md
```diff
@@ -116,3 +116,34 @@ when a share is sent to a native application, the user agent may create an
...
+  third-party application that does not respect the user's privacy setting.
+  Implementations should consider presenting additional warnings or disabling
+  the feature entirely when in a private browsing mode.
+* The data passed to `navigator.share` may be used to exploit buffer overflow
```
marcoscaceres: as above... probably can.

docs/interface.md
mgiuca: In that case, I will add `[SecureContext]` to the WebIDL instead.

docs/interface.md
mgiuca: > s/which apps/which native applications

No, this advice should not only be specific to native applications. If we allow websites as the target, then user agents should also not allow the website to learn which websites have been added or chosen.

docs/interface.md
mgiuca: I have been deliberately avoiding going too much into spec language, as I think it makes it less understandable. We can go to that when we go into a proper spec, but for now, I want to keep it casual and readable to someone who doesn't read web specs. I'd rather just use an example here rather than linking to spec documents.

docs/interface.md
mgiuca: As above, I'm not too concerned about getting the exact terminology right. I think "may" (not capitalized) is more appropriate here, and in the other places -- I want to highlight that this is a possibility, not a certainty.

docs/interface.md
mgiuca: Especially here: "may be used" indicates that it's a possibility. "can be used" means that it is able to be exploited, which is not necessarily true.

docs/interface.md
marcoscaceres: ok, that's fair.

docs/interface.md
marcoscaceres: sgtm.

docs/interface.md
marcoscaceres: ah yeah, I should have have said "could be used" - but what is there is fine.

PR #24, opened 2017-05-02
Adds Web Share spec draft

mgiuca:
Readable copy of this draft on [RawGit](https://rawgit.com/mgiuca/web-share/spec/index.html).

Marcos, perhaps you could take a look at this draft, when you have time. I've had Shane Stephens at Google take a first pass over it and made some changes he suggested.

### DIFF
--- a/docs/interface.md
+++ b/docs/interface.md
@@ -1,12 +1,14 @@
 # Web Share API Interface
 
-**Date**: 2016-05-30
+**Date**: 2017-05-02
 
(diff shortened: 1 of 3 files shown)


mgiuca:
Lol, I think I never added Marcos (GitHub UI, so confusing).

Well, in any case, since there's nothing there at the moment I think it's simpler if I just land the draft I have, then we can take further discussions as issues. (Easier than doing it in one big code review.)

PR #26, opened 2017-05-31
Correct metadata for being a WICG document.

dbaron:
This also uses the github key rather than manually (and differently) generating the links that including it generates.


dbaron:
(I got here from w3ctag/design-reviews#179.)

mgiuca:
Great, thanks for updating this. I wasn't sure how to fix a lot of those problems up the top.

PR #27, opened 2017-05-31
Spell check.

dbaron:


mgiuca:
Thanks (though "Acknowledgement" is correct British English and "advocation" is apparently valid "old English"), I'll allow it :)

Issue #29, opened 2017-06-02
Don't allow the caller to distinguish different failure modes


mgiuca:
Raised by @dbaron on the [TAG review](https://github.com/w3ctag/design-reviews/issues/179#issuecomment-305326710).

There is a minor privacy issue in distinguishing between "no apps are available" and "the user cancelled the share"; if only a handful of popular targets exist for a particular payload (especially if we add new share types that might have specialized handlers, e.g., for rare MIME types) then a malicious site could use `navigator.share` to check whether those apps are installed, based on the response string.

Also, in our current Android implementation, we are totally unable to distinguish between "no apps are available" and "the user cancelled the share". Therefore, let's just remove this provision. I won't go as far as to explicitly bar the user agent from distinguishing these, but add a note under security considerations.

PR #30, opened 2017-06-02
Share method: Remove the ability for the share target to decline.

### LINKED ISSUE #28: Don't allow the target to decline a share
The current spec has a provision for the target to "[**decline to receive the data**](https://wicg.github.io/web-share/#dfn-decline-to-receive-the-data)". Despite this, neither the existing Android implementation on Chrome nor the proposed [Web Share Target](https://github.com/WICG/web-share-target) URL-template-based API would allow the target to respond with a failure state.

Given that we aren't likely to make use of this in the near future, and that it could be re-added in a backwards-compatible way if we wanted to introduce a new failure mode, let's just remove this text from the spec for now.

mgiuca:
Closes #28.

### DIFF
--- a/index.html
+++ b/index.html
@@ -140,8 +140,7 @@
                 <var>data</var> to a format suitable for ingestion into the
                 target</a>, and transmit the converted data to the target. If
                 an error occurs starting the target or transmitting the data,
-                or if the target <a>declines to receive the data</a>, reject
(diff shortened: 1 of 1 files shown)


mgiuca:
@sammc PTAL since you suggested this change.

Also note follow-up #29 and 25b8756c.

PR #31, opened 2017-06-02
Removed note about distinguishing between different error cases.

### LINKED ISSUE #29: Don't allow the caller to distinguish different failure modes

mgiuca:
Added a security consideration that implementations should be careful
about exposing this information.

This isn't a technical change, but changes a weak recommendation *to*
differentiate error cases to a weak recommendation *against* it.

Closes #29.

### DIFF
--- a/index.html
+++ b/index.html
@@ -161,13 +161,6 @@
             perform any kind of "always use this target" to bypass the UI in
             subsequent share operations.
           </div>
-          <div class="note">
-            The user agent MAY differentiate between the three different causes
-            of <a data-cite="!WEBIDL#aborterror"><code>AbortError</code></a>
-            failure (e.g., with a different error string in each case).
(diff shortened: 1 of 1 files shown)


mgiuca:
@sammc What do you think of this one?

PR #35, opened 2017-06-16
spec: share now resolves url relative to the document base URL.

### LINKED ISSUE #34: Spec: URL is relative to current page and converted to absolute
Current behaviour in Chrome (intentional) that wasn't captured in the spec: the "`url`" field can be a relative URL, and if it is, it is converted into an absolute URL based on the current location, in the same way as relative URLs in links are.

In particular, sharing a URL of `'.'` shares the current page location's directory (without the final path segment); sharing a URL of `''` shares the current URL.

mgiuca:
This is consistent with the current implementation in Chrome.

Closes #34.

***
[Preview](https://s3.amazonaws.com/pr-preview/mgiuca/web-share/spec-url-relative.html) | [Diff](https://s3.amazonaws.com/pr-preview/WICG/web-share/dac4c21...mgiuca:69f29d6.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -66,9 +66,9 @@
         });
       </pre>
       <p>
-        In response to this call to <code>navigator.share</code>, the user
-        agent would display a picker or chooser dialog, allowing the user to
-        select a target to share this title and URL to.
+        In response to this call to <a><code>navigator.share</code></a>, the
+        user agent would display a picker or chooser dialog, allowing the user
+        to select a target to share this title and URL to.
       </p>
     </section>
     <section>
--- a/index.html
+++ b/index.html
@@ -112,6 +112,17 @@
           <ol>
             <li>Let <var>p</var> be a newly created promise.
             </li>
+            <li>Let <var>url</var> be the result of running the <a data-cite=
+            "!URL#concept-url-parser">URL parser</a> on <var>data</var>'s
+            <a for="ShareData">url</a>, with the document's <a data-cite=
+            "!HTML#document-base-url">base URL</a>, and no <var>encoding
+            override</var>.
+            </li>
+            <li>Set <var>data</var> to a copy of <var>data</var>, with its
+            <a for="ShareData">url</a> field set to the result of running the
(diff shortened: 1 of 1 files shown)


mgiuca:
@sammc can you look at this?

Note there's a nifty new "Preview" and "Diff" link in the PR which you can use to see the delta on the rendered HTML (but the example code is a bit borked in the Diff view).

index.html
```diff
@@ -118,6 +123,17 @@
             "!WEBIDL#securityerror"><code>SecurityError</code></a>, and abort
             these steps.
             </li>
+            <li>Let <var>url</var> be the result of running the <a data-cite=
```
sammc: What if the URL isn't valid?

mgiuca: That's a good point. I thought the URL parser was resilient to invalid URLs but it seems it has two levels of failure, "validation error" (a warning which is generally ignored) and "failure" (a parsing error).

One such fatal error is an invalid port number:

```js
navigator.share({url: 'http://localhost:65536'});
```

In Chrome, this currently fails silently, replacing the URL with the empty string and completing the successful share. Do you think this should cause navigator.share to fail with a new error? (I do.)

mgiuca: And if so, I think it should be a [TypeError](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror), which is apparently the JavaScript equivalent of Python's ValueError and TypeError combined, not a [SyntaxError](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-syntaxerror) which is reserved for JavaScript parse errors.

mgiuca: FYI: I've filed a Chrome bug [crbug.com/734486](https://crbug.com/734486) for updating Chrome's behaviour.

sammc: TypeError sounds good. It matches [fetch](https://fetch.spec.whatwg.org/#request-class).

mgiuca: Done, PTAL.

PR #42, opened 2017-06-20
spec: Check whether url is present before processing it.

### LINKED ISSUE #39: Spec bug: Deal with undefined or missing url
Currently if you do `navigator.share()` on `https://example.com/`, it will share `https://example.com/undefined`, by my reading. (I.e. it will convert `undefined` to a `USVString`, then resolve it as a relative URL.)

If you add a default to the dictionary of `url = ""`, it will instead share the current page URL, which seems nice.

mgiuca:
Fixes a spec bug where a non-existent URL would be parsed.

Closes #39.

***
[Preview](https://s3.amazonaws.com/pr-preview/mgiuca/web-share/spec-check-missing-url.html) | [Diff](https://s3.amazonaws.com/pr-preview/WICG/web-share/0be490e...mgiuca:3386eb5.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -117,21 +117,26 @@
           <ol>
             <li>Let <var>p</var> be a newly created promise.
             </li>
-            <li>Let <var>url</var> be the result of running the <a data-cite=
-            "!URL#concept-url-parser">URL parser</a> on <var>data</var>'s
-            <a for="ShareData">url</a>, with the document's <a data-cite=
-            "!HTML#document-base-url">base URL</a>, and no <var>encoding
-            override</var>.
-            </li>
-            <li>If <var>url</var> is failure, reject <var>p</var> with
-            <a data-cite=
-            "!WEBIDL#exceptiondef-typeerror"><code>TypeError</code></a>, and
-            abort these steps.
-            </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -117,21 +117,26 @@
...
-            abort these steps.
-            </li>
-            <li>Set <var>data</var> to a copy of <var>data</var>, with its
-            <a for="ShareData">url</a> field set to the result of running the
-            <a data-cite="!URL#concept-url-serializer">URL serializer</a> on
-            <var>url</var>.
+            <li>If <var>data</var>'s <a for="ShareData">url</a> member is
+            <a data-cite="!WEBIDL#dfn-present">present</a>:
+              <ol>
+                <li>Let <var>url</var> be the result of running the
```
mgiuca: Everything below this is unchanged (just indented).

mgiuca:
@domenic This should fix #39. As I explained on that bug, I don't want to make it default to the empty string. This simple change matches the behaviour in Chrome which is to just have no url attribute by default.

mgiuca:
(@domenic has lgtm'd this on #39)

Issue #43, opened 2017-06-21
Spec: Resolve p with undefined


mgiuca:
Minor nit pointed out by @foolip [here](https://github.com/w3c/web-platform-tests/pull/6287#discussion_r123085848) --- you can't just "resolve" a promise, even if it's void, need to resolve "with undefined".

Issue #40, opened 2017-06-20
Non-normative security and privacy section contains normative RFC 2119 keywords


domenic:
It seems like the section should just be marked normative

mgiuca:
I see, so you should not have those keywords in non-normative text at all?

Most of the security section is just explanations of other normative requirements, as well as considerations that can't be formalized as requirements (such as "think about private browsing modes"). I don't think it makes sense to make this section normative.

Instead, I will move that MUST NOT requirement (leaking info about installed apps) into a normative section, and rewrite the security section to be an explanation of those requirements, rather than a statement. There's a similar thing in a NOTE just under the share algorithm which will get the same treatment.

domenic:
> I see, so you should not have those keywords in non-normative text at all?

Yeah, it's a bug to state normative requirements in non-normative text. If you're depending on RFC 2119 (either directly, or indirectly via Infra) then any occurrences of those words are to be interpreted as described there, so there's no getting around it.

As you can see from my review of #44 there's a bit of an art to using other phrases (like "can" or "might") instead of these "reserved keywords".

PR #44, opened 2017-06-21
Miscellaneous spec cleanup (no normative changes)

### LINKED ISSUE #36: "the document's URL" is ambiguous
Consider the situation:

```js
function declaredInWindow3() {
  window1.navigator.share.call(window2.navigator, { url });
}
```

Then navigate the frame containing window3 to window4, but save a reference to `declaredInWindow3`. Finally, from code in window5, call the saved function.

Which of these five documents does `url` get resolved relative to?

The proper pattern for this is to parse against

> `this Navigator object`'s [relevant settings object](https://html.spec.whatwg.org/multipage/webappapis.html#relevant-settings-object)'s [API base URL](https://html.spec.whatwg.org/multipage/webappapis.html#api-base-url)

in which case the answer will be window2.

### LINKED ISSUE #37: Editorial: various missing links
- URL parser: should go to https://url.spec.whatwg.org/#concept-url-parser (but see #38)
- Triggered by user activation: https://html.spec.whatwg.org/#triggered-by-user-activation
- In parallel: https://html.spec.whatwg.org/#in-parallel
- Valid URL string: https://url.spec.whatwg.org/#valid-url-string
- Promise creation/manipulation should link to https://www.w3.org/2001/tag/doc/promises-guide
- Exceptions should link to Web IDL (and also should use the somewhat-new syntax, e.g. "AbortError" DOMException instead of just AbortError, since there is no type called AbortError).

In Bikeshed, most of this should be a simple matter of e.g. doing `[=in parallel=]` instead of `in parallel`.

mgiuca:
Closes #36, #37, #40 and #43.

***
[Preview](https://s3.amazonaws.com/pr-preview/mgiuca/web-share/spec-cleanup.html) | [Diff](https://s3.amazonaws.com/pr-preview/WICG/web-share/857fe5c...mgiuca:b658e25.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -127,7 +127,8 @@
                   <a data-cite="!HTML#document-base-url">base URL</a>, and no
                   <var>encoding override</var>.
                 </li>
-                <li>If <var>url</var> is failure, reject <var>p</var> with
+                <li>If <var>url</var> is failure, <a data-cite=
+                "!promises-guide#reject-promise">reject</a> <var>p</var> with
                 <a data-cite=
                 "!WEBIDL#exceptiondef-typeerror"><code>TypeError</code></a>,
                 and abort these steps.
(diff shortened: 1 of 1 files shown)


mgiuca:
@domenic Fixes #37 and #40 that you filed, looks good?

index.html
```diff
@@ -112,10 +112,11 @@
           </h4>
           <p>
             When the <dfn>share</dfn> method is called with argument
-            <var>data</var>, run the following steps:
+            <var>data</var>, the user agent MUST run the following steps:
```
domenic: FYI at least in specs that depend on Infra, this is implied: https://infra.spec.whatwg.org/#algorithms

index.html
```diff
@@ -330,22 +335,23 @@
...
-        installed, or which app was chosen from
-        <a><code>navigator.share</code></a>. This information could be used for
-        fingerprinting, as well as leaking details about the user's device.
+        <li>There is a requirement to not allow the website to learn which apps
+        are installed, or which app was chosen from
+        <a><code>navigator.share</code></a>, because this information could be
+        used for fingerprinting, as well as leaking details about the user's
+        device.
         </li>
         <li>Implementors should carefully consider what information is revealed
```
domenic: "should" is still normative; maybe move this to the step that rejects the promise?

index.html
```diff
@@ -330,22 +335,23 @@
...
-        security confirmation, ensuring that websites cannot silently send data
-        to native applications.
+        <li>There is a requirement that <a><code>navigator.share</code></a>
+        presents the user with a dialog asking them to select a target
+        application (even if there is only one possible target). This surface
+        serves as a security confirmation, ensuring that websites cannot
+        silently send data to native applications.
         </li>
         <li>Due to the capabilities of the API surface,
         <a><code>navigator.share</code></a> is <a data-cite=
```
domenic: There's an errant "may leak private data" below that ideally would be "might". The same bullet also contains a "should" which I am not sure what to do with.

Similarly, "may be used" should be "can be used", and "should be aware" should be, e.g., "will want to be aware"

index.html
```diff
@@ -330,22 +335,23 @@
         (depending on the underlying platform).
       </p>
       <ul>
```
domenic: The introduction still contains "should", and seems out of place since this is supposed to be now re-stating requirements expressed elsewhere.

index.html
```diff
@@ -171,21 +172,25 @@
                 these steps.
```
domenic: Above this, "reject" should also link to the promises-guide

mgiuca: Done.

index.html
mgiuca: Done. I've grepped the whole document for all reserved words in RFC 2119 and rewritten the sentences around them (or, in one case, upgraded "may" to "MAY").

index.html
mgiuca: Done.

"will want to be aware" sounds very watery, but I can't think of anything better that doesn't involve these words.

mgiuca:
PTAL.

mgiuca:
Added a fix for #36 as well. PTAL.

PR #49, opened 2017-07-03
spec: Changed user gesture error from SecurityError to NotAllowedError.

### LINKED ISSUE #45: User gesture check should fail with some other error type (not SecurityError)?
@foolip reported [here](https://github.com/w3c/web-platform-tests/pull/6287#discussion_r123089862):

> ...if you feel like changing it, I'd argue that user gesture checks are not quite security sensitive, but rather prevent user annoyance. FWIW, https://fullscreen.spec.whatwg.org/ rejects with TypeError.

@sammc I think we borrowed the SecurityError from some other API, so there may be two different precedents. Do you remember which?

mgiuca:
Minor breaking change, as discussed on #45. Closes #45.

***
[Preview](https://s3.amazonaws.com/pr-preview/mgiuca/web-share/gesture-error-type.html) | [Diff](https://s3.amazonaws.com/pr-preview/WICG/web-share/5c2c8bb...mgiuca:eda2322.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -149,8 +149,8 @@
             activation</a>, <a data-cite=
             "!promises-guide#reject-promise">reject</a> <var>p</var> with
             <a data-cite=
-            "!WEBIDL#securityerror"><code>SecurityError</code></a>, and abort
-            these steps.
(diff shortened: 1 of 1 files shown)


mgiuca:
@sammc ptal.

PR #50, opened 2017-07-03
spec: Throw a TypeError if all fields are missing.

### LINKED ISSUE #48: navigator.share() should throw a TypeError if all fields are omitted
The following call:

> navigator.share({foo: 'xyz'})

Currently results in a successful share with no data.

It should reject with a `TypeError`.

Rationale: Empty shares are a bad user experience. They show up in the target application as empty strings which is confusing. There's the simple case of passing an empty dictionary to `share`, but we also have to consider the more plausible case of adding new fields in the future, and how sites that use the new API will interact with old UAs that don't support those new fields.

The current status quo is that if you share *only* new fields, you'll share nothing on an old UA.

We should offer feature detection for these new fields (see [WebIDL #107](https://github.com/heycam/webidl/issues/107)) and then make it a TypeError if you fail to properly use the feature detection and only share new fields.

This is related to #47 which is about making `navigator.share()` equivalent to `navigator.share({})`. If we make this change, `navigator.share()` would remain a TypeError which makes me happy. I think `navigator.share()` being a successful share is nonsensical.

mgiuca:
This makes sharing the empty dictionary, or a dictionary with only unrecognised fields, a programmer error, and is designed to aid in future extensions (since if new fields are added, it will be a TypeError to exclusively use them without feature detection).

Also adds an appendix discussing extending the spec in the future.

Closes #48.

***
[Preview](https://s3.amazonaws.com/pr-preview/mgiuca/web-share/empty-dict-error.html) | [Diff](https://s3.amazonaws.com/pr-preview/WICG/web-share/5c2c8bb...mgiuca:0c7a74d.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -388,6 +388,60 @@
         </li>
       </ul>
     </section>
+    <section class="appendix informative">
+      <h2>
+        Extensibility of this API
+      </h2>
+      <p>
+        The Web Share API is designed to be extended in the future by way of
+        new members added to to the <a>ShareData</a> dictionary, to allow both
+        sharing of new types of data (<i>e.g.</i>, <a href=
+        "https://github.com/WICG/web-share/issues/12">images</a>) and strings
+        with new semantics (<i>e.g.</i> author).
+      </p>
+      <div class="warning">
+        This doesn't mean user agents can add whatever members they like. It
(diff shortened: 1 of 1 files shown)


mgiuca:
@sammc ptal.

PR #51, opened 2017-07-04
spec: Make ShareData argument optional.

### LINKED ISSUE #47: navigator.share's argument needs to be optional because ShareData has no required members
This is another thing that's a bit surprising in Web IDL, but if a dictionary has no required members, then when used as an argument type, that argument must be optional. It's this bit from https://heycam.github.io/webidl/#idl-operations:
> If the type of an argument is a dictionary type or a union type that has a dictionary as one of its flattened member types, and that dictionary type and its ancestors have no required members, and the argument is either the final argument or is followed only by optional arguments, then the argument must be specified as optional. Such arguments are always considered to have a default value of an empty dictionary, unless otherwise specified.

The reason is that with no required arguments, `navigator.share({})` still works, and then it looks "silly" to require the empty object.

mgiuca:
This has no behavioural change, as passing an empty dictionary is a
TypeError. But it is required by the WebIDL spec.

Closes #47.

***
[Preview](https://s3.amazonaws.com/pr-preview/mgiuca/web-share/optional-argument.html) | [Diff](https://s3.amazonaws.com/pr-preview/WICG/web-share/4e65307...mgiuca:f421675.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -90,9 +90,19 @@
         </p>
         <pre class="idl">
           partial interface Navigator {
-            [SecureContext] Promise&lt;void&gt; share(ShareData data);
+            [SecureContext] Promise&lt;void&gt; share(optional ShareData data);
           };
         </pre>
+        <div class="note">
+          The <var>data</var> argument is marked as "optional", but it is
+          effectively required; <a for="Navigator"><code>share</code></a> will
+          throw a <a data-cite=
+          "!WEBIDL#exceptiondef-typeerror"><code>TypeError</code></a> if called
+          with no arguments. WebIDL <a data-cite=
+          "!WEBIDL#idl-operations">mandates that it be optional</a> because the
+          dictionary members are all optional. See WebIDL <a href=
+          "https://github.com/heycam/webidl/issues/130">Issue #130</a>.
+        </div>
         <p>
           User agents that do not support sharing SHOULD NOT expose <a for=
           "Navigator"><code>share</code></a> on the


mgiuca:
@sammc ptal.

index.html
```diff
@@ -90,9 +90,19 @@
...
+            [SecureContext] Promise&lt;void&gt; share(optional ShareData data);
           };
         </pre>
+        <div class="note">
+          The <var>data</var> argument is marked as "optional", but really it
+          is required; <a for="Navigator"><code>share</code></a> will throw a
+          <a data-cite=
+          "!WEBIDL#exceptiondef-typeerror"><code>TypeError</code></a> if called
+          with no arguments. WebIDL <a data-cite=
+          "!WEBIDL#idl-operations">mandates that it be optional</a>, because
```
sammc: I don't think you want a comma here.

index.html
```diff
@@ -90,9 +90,19 @@
         </p>
         <pre class="idl">
           partial interface Navigator {
-            [SecureContext] Promise&lt;void&gt; share(ShareData data);
+            [SecureContext] Promise&lt;void&gt; share(optional ShareData data);
           };
         </pre>
+        <div class="note">
+          The <var>data</var> argument is marked as "optional", but really it
```
sammc: The really feels a bit superfluous.

index.html
mgiuca: Done.

index.html
mgiuca: Shane suggested I write a comment explaining why this is here, since it's quite unintuitive to have an optional argument that is mandatory.

In particular, I think while developers may not read the whole spec, the one thing that is fairly understandable to a non-web-spec person is IDL, so we may have confused developers reading the IDL and wondering why this is "optional". That's why I'm so strongly opposed to adding a misleading "optional" here, but if we are going to have it, I'd rather explain it.

If you'd like I can reduce the amount of text here and perhaps convert it into a `//` comment in the IDL itself rather than a note.

sammc: I meant just the word "really." The rest is fine.

mgiuca: Oh lol, use quotes next time. (I read it as "This really feel a bit superfluous".)

PR #56, opened 2017-11-27
Modernize example and show that it throws

marcoscaceres:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/web-share/pull/56.html" title="Last updated on Feb 12, 2018, 3:05 AM GMT (16f99af)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/web-share/56/d39a1a7...16f99af.html" title="Last updated on Feb 12, 2018, 3:05 AM GMT (16f99af)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -65,7 +65,7 @@
             await navigator.share({ title: "Example Page", url: "" });
             console.log("Share successful");
           } catch (err) {
-            console.error("Oh no! couldn't share.", err.message);
+            console.error("Share failed:", err.message);
           }
         });
       </pre>
--- a/index.html
+++ b/index.html
@@ -60,12 +60,12 @@
         click, this JavaScript code shares the current page's URL.
       </p>
       <pre class="example javascript" title="Basic usage">
-        shareButton.addEventListener('click', async () =&gt; {
+        shareButton.addEventListener("click", async () => {
           try {
-            await navigator.share({title: 'Example Page', url: ''});
-            console.log('Share successful');
+            await navigator.share({ title: "Example Page", url: "" });
+            console.log("Share successful");
           } catch (err) {
-            console.error('Oh no! couldn't share. ' + err.message);
+            console.error("Oh no! couldn't share.", err.message);
           }
         });
       </pre>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Single quotes are the devil :)

maxnordlund:
Does it really throw? Seems like it returns a promise, and only because your example uses async/await it turns it into a throw, right?

Anyway, having both a promise based, and async/await based example is nice.

index.html
```diff
@@ -60,9 +60,13 @@ <h2>
...
       <pre class="example javascript" title="Basic usage">
-        shareButton.addEventListener('click', () =&gt; {
-          navigator.share({title: 'Example Page', url: ''})
-              .then(console.log('Share successful'));
+        shareButton.addEventListener('click', async () =&gt; {
+          try {
+            await navigator.share({title: 'Example Page', url: ''});
+            console.log('Share successful');
+          } catch (err) {
+            console.error('Oh no! couldn't share. ' + err.message);
```
ericwilligers: The single quote in `couldn't` needs escaping.

marcoscaceres:
> Does it really throw? Seems like it returns a promise, and only because your example uses async/await it turns it into a throw, right?

Correct.

> Anyway, having both a promise based, and async/await based example is nice.

Given async/await, using raw promise (particularly examples that use `.then()` and `.catch()`) is a bit of an anti-pattern - so I'd like us not to just stick to async/await examples. Keeps things much cleaner. We should only revert back to `.then`, `.catch()`, etc. if there is a valid reason to do so.

index.html
```diff
@@ -60,9 +60,13 @@ <h2>
...
       <pre class="example javascript" title="Basic usage">
-        shareButton.addEventListener('click', () =&gt; {
-          navigator.share({title: 'Example Page', url: ''})
-              .then(console.log('Share successful'));
+        shareButton.addEventListener("click", async () => {
+          try {
+            await navigator.share({ title: "Example Page", url: "" });
+            console.log("Share successful");
+          } catch (err) {
+            console.error("Oh no! couldn't share.", err.message);
```
mgiuca: I don't like this informal message. Can you change the string to "Share failed:"

PR #72, opened 2018-02-20
Extract 'Extensibility of the API'

### LINKED ISSUE #61: Extensibility of the API section could be removed
I'm not a huge fan of "Extensibility of the API" section, tbh. I don't think we should include it: The Share API is not any more special wrt extensibility than any other API. Same extensibility rules apply to platform APIs.

ewilligers:
The chapter 'Extensibility of the API' is moved into a separate
document. It remains available as a resource for informing future
evolution of the specification, without itself being part of the
specification.

resolves #61

***
<a href="https://pr-preview.s3.amazonaws.com/ewilligers/web-share/pull/72.html" title="Last updated on Feb 20, 2018, 3:14 AM GMT (5214ef1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/web-share/72/62ffe50...ewilligers:5214ef1.html" title="Last updated on Feb 20, 2018, 3:14 AM GMT (5214ef1)">Diff</a>

### DIFF
--- a/extending.html
+++ b/extending.html
@@ -1,0 +1,79 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>
+      Extending the Web Share API
+    </title>
+    <script src="https://www.w3.org/Tools/respec/respec-w3c-common" class=
+    "remove"></script>
+  </head>
+  <body>
+    <section style="display:none;">
+      <h2>
+        Dependencies
+      </h2>
+      <p>
+        <code><dfn data-cite=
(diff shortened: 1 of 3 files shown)


extending.html
```diff
@@ -0,0 +1,79 @@
+<!DOCTYPE html>
```
mgiuca: If this is going to go in a separate, informal, document, then:

1. It should go in /docs.
2. I think it should just be a plain Markdown file (like the other informal docs in that directory), rather than a Respec-formatted doc.

PR #78, opened 2018-09-26
Sharing files

ewilligers:
Add a files member to Web Share dictionary.

Add canShare() so authors can determine if an implementation supports sharing a given ShareData dictionary. In particular, this allows authors to determine if an implementation supports sharing files.

***
<a href="https://pr-preview.s3.amazonaws.com/ewilligers/web-share/pull/78.html" title="Last updated on Oct 10, 2018, 12:25 AM GMT (18e0eff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/web-share/78/05a87d5...ewilligers:18e0eff.html" title="Last updated on Oct 10, 2018, 12:25 AM GMT (18e0eff)">Diff</a>

### DIFF
--- a/level-2/index.html
+++ b/level-2/index.html
@@ -100,6 +100,7 @@
         </p>
         <pre class="idl">
           partial interface Navigator {
+            [SecureContext] boolean canShare(optional ShareData data);
             [SecureContext] Promise&lt;void&gt; share(optional ShareData data);
           };
         </pre>
@@ -125,6 +126,17 @@
         </div>
         <section>
           <h4>
+            <dfn>canShare()</dfn> method
+          </h4>
+          <p>
+            <code>canShare(</code><var>data</var><code>)</code> MUST return
+            <code>true</code> unless
+            <code>share(</code><var>data</var><code>)</code> would reject with
+            <a>TypeError</a>, in which case it MUST return <code>false</code>.
+          </p>
+        </section>
+        <section>
+          <h4>
             <dfn>share()</dfn> method
           </h4>
           <p>
@@ -136,7 +148,8 @@
             "!promises-guide#a-new-promise">a new promise</a>.
             </li>
             <li>If none of <var>data</var>'s members <a>title</a>, <a>text</a>,
-            or <a>url</a> are present, reject <var>p</var> with
+            <a>url</a> or (if the implementation supports file sharing)
+            <a>files</a> are present, reject <var>p</var> with
             <a>TypeError</a>, and abort these steps.
             </li>
             <li>If <var>data</var>'s <a>url</a> member is <a data-cite=
@@ -218,6 +231,11 @@
             kind of "always use this target in the future" option, or bypassing
             the UI if there is only a single share target.
           </div>
+          <div class="note">
+            To determine if file sharing is supported by an implementation,
+            <a>canShare()</a> can be called with a dictionary containing only
+            <code>files</code>.
+          </div>
         </section>
       </section>
       <section data-dfn-for="ShareData" data-link-for="ShareData">
@@ -229,6 +247,7 @@
             USVString title;
             USVString text;
             USVString url;
+            FileList files;
           };
         </pre>
         <p>
@@ -255,9 +274,17 @@
           <dd>
             A URL string referring to a resource being shared.
           </dd>
+          <dt>
+            <dfn>files</dfn> member
+          </dt>
+          <dd>
+            A <code><dfn data-cite=
+            "!FILE-API#filelist-section">FileList</dfn></code> referring to one
+            or more files being shared.
+          </dd>
         </dl>
         <div class="note">
-          These members are <a data-cite=
+          The string members are <a data-cite=
           "!WEBIDL#idl-USVString"><code>USVString</code></a> (as opposed to
           <a data-cite="!WEBIDL#idl-DOMString"><code>DOMString</code></a>)
           because they are not allowed to contain invalid <a data-cite=


index.html
```diff
@@ -224,6 +243,7 @@ <h3>
             USVString title;
             USVString text;
             USVString url;
+            FileList files;
```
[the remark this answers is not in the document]
mgiuca: Also, I wonder if `FileList` should be considered deprecated, due to the comment [here](https://www.w3.org/TR/FileAPI/#filelist-section):

> Note: The FileList interface should be considered "at risk" since the general trend on the Web Platform is to replace such interfaces with the Array platform object in ECMAScript [ECMA-262].

index.html
```diff
@@ -95,6 +95,7 @@ <h3>
         </p>
         <pre class="idl">
           partial interface Navigator {
+            [SecureContext] boolean canShare(optional ShareData data);
```
mgiuca: We've talked on heycam/webidl#107 about a more generic solution to feature detection of dictionary members, than every API making its own `canX` method. Perhaps we should check up on the status of that.

index.html
```diff
@@ -119,6 +120,24 @@ <h3>
...
         <section>
+          <h4>
+            <dfn>canShare()</dfn> method
+          </h4>
+          <p>
+            When the <a>canShare()</a> method is called with argument
+            <var>data</var>, run the following steps:
+          </p>
+          <ol data-link-for="ShareData">
+            <li>If none of <var>data</var>'s members <a>title</a>, <a>text</a>,
```
mgiuca: This doesn't really capture what happens if `files` is not supported. I know the spec assumes that it is (since an implementation compliant with the latest spec would support file sharing), but what if it isn't?

If files isn't supported, does it return true or false if I include both `files` and some other supported field?

ewilligers: Additional unsupported dictionary fields are ignored. (We could mention that explicitly.) If an implementation chose to support canShare but not file sharing, then (a) that implementation would not be consistent with any version of the spec, (b) that implementation should return false if the dictionary only contained files. If would return true if the dictionary also contained some other supported field.

ewilligers: Adding text to say the method must return false if the dictionary contains `files` and file sharing is not supported would be possible for the special case of `files`, but similar text would probably not be possible for any future fields that might be added in future versions.

index.html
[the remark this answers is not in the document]
ewilligers: The link in the IDL is now fixed.

index.html
mgiuca: > If an implementation chose to support canShare but not file sharing, then (a) that implementation would not be consistent with any version of the spec

I guess this is important for how developers are expected to use it.

Are we leaving open a possibility of an intermediate browser that supports `canShare` but not files? It's possible that a future browser will never want to support file sharing (because the OS doesn't have a concept of files, or whatever), but will want to support some other new feature that requires `canShare`. So I think we should allow this, possibly even explicitly saying that files is a MAY (not a MUST), and linking up `canShare` so it MUST return true if and only if files is accepted.

If we are allowing this possibility of `canShare` without `files`, then we need to enumerate this case, and that means developers are expected to check for both `canShare` and `files`, like this:

```js
if (navigator.share && navigator.canShare && navigator.canShare({files})) {
  await showShareButton();
  navigator.share({files});
}
```

If we aren't, then it means the very existence of `canShare` implies that `files` is supported, so then developers have a legitimate expectation of this code always being correct:

```js
if (navigator.share && navigator.canShare) {
  await showShareButton();
  navigator.share({files});
}
```

The latter option seems silly to me, so I think we should require developers to actually call `canShare` if they want to check if file sharing is allowed, which means we need to allow `canShare` to exist, but return false if given `files`.

index.html
```diff
@@ -119,6 +120,31 @@ <h3>
...
+            <a>url</a> or <a>files</a> are present, return <code>false</code>.
+            </li>
+            <li>return <code>true</code>.
+            </li>
+          </ol>
+          <div class="note" data-link-for="ShareData">
+            <p>
+              <code>canShare()</code> MUST return <code>false</code> whenever
```
mgiuca: This is normative, and therefore shouldn't be in a note. (I must've missed this text last time I looked.)

This is actually not true, as stated. If `url` is invalid, `share` should reject with `TypeError`, but `canShare` should return true because it doesn't check URL. I think `canShare` *should* return false if `url` is invalid, and thus this statement here better matches what I want than the definition above.

I think I'd be happier with this being the *definition* of `canShare`, rather than a corollary (so we don't have to duplicate or abstract out the checking logic). Do you think we can get away with simply deleting the steps, and writing "`canShare()` MUST return false whenever ..... It MUST return true otherwise." ?

index.html
mgiuca: Discussion is ongoing...

index.html
```diff
@@ -118,6 +119,17 @@ <h3>
           Clients can use the presence or absence of this method to determine
           whether to show UI that triggers its use.
         </div>
+        <section>
+          <h4>
+            <dfn>canShare()</dfn> method
+          </h4>
+          <p>
+            <code>canShare(</code><var>data</var><code>)</code> returns
```
mgiuca: MUST return true unless .... in which case it MUST return false.

index.html
```diff
@@ -131,9 +143,14 @@ <h4>
             "!promises-guide#a-new-promise">a new promise</a>.
             </li>
             <li>If none of <var>data</var>'s members <a>title</a>, <a>text</a>,
-            or <a>url</a> are present, reject <var>p</var> with
+            <a>url</a> or <a>files</a> are present, reject <var>p</var> with
             <a>TypeError</a>, and abort these steps.
             </li>
+            <li>If the implementation does not support file sharing and none of
```
mgiuca: This does not seem sustainable; imagine adding another optional member, the permutations.

I would rather just delete this, and add a parenthetical above, like this:

"If none of data's members title, text, or url are present, reject p with url or (if the implementation supports file sharing) files are present"

PR #85, opened 2018-12-20
Optional message to display when sharing

ewilligers:
For sites that wish to display a message when the
share dialog is presented, we add a new ShareOptions
dictionary.

This allows for additional options that may be needed in future,
without long method parameter lists.

Note that support for the 'message' option (and likely future
options) is not feature detectable.

***
<a href="https://pr-preview.s3.amazonaws.com/ewilligers/web-share/pull/85.html" title="Last updated on Dec 21, 2018, 4:17 AM UTC (4d026ec)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/web-share/85/d9d43d7...ewilligers:4d026ec.html" title="Last updated on Dec 21, 2018, 4:17 AM UTC (4d026ec)">Diff</a>

### DIFF
--- a/level-2/index.html
+++ b/level-2/index.html
@@ -101,8 +101,8 @@
         </p>
         <pre class="idl">
           partial interface Navigator {
-            [SecureContext] boolean canShare(optional ShareData data);
-            [SecureContext] Promise&lt;void&gt; share(optional ShareData data);
+            [SecureContext] boolean canShare(optional ShareData data, optional ShareOptions options);
+            [SecureContext] Promise&lt;void&gt; share(optional ShareData data, optional ShareOptions options);
           };
         </pre>
         <div class="note">
@@ -130,9 +130,10 @@
             <dfn>canShare()</dfn> method
           </h4>
           <p>
-            <code>canShare(</code><var>data</var><code>)</code> MUST return
-            <code>true</code> unless
-            <code>share(</code><var>data</var><code>)</code> would reject with
+            <code>canShare(</code><var>data</var><code>,</code>
+            <var>options</var><code>)</code> MUST return <code>true</code>
+            unless <code>share(</code><var>data</var><code>,</code>
+            <var>options</var><code>)</code> would reject with
             <a>TypeError</a>, in which case it MUST return <code>false</code>.
           </p>
         </section>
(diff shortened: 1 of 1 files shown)


level-2/index.html
```diff
@@ -130,9 +130,10 @@ <h4>
             <dfn>canShare()</dfn> method
           </h4>
           <p>
-            <code>canShare(</code><var>data</var><code>)</code> MUST return
-            <code>true</code> unless
-            <code>share(</code><var>data</var><code>)</code> would reject with
+            <code>canShare(</code><var>data</var><code>,</code>
```
mgiuca: What are your thoughts on making `canShare` not take `options` (since, at least for now, there are no options that should affect the shareability of the data).

I could go either way on this.

ewilligers: That works.

We could make `canShare(null, {message: ""})` be a way to detect if `message` is supported, but I don't think it is worthwhile.

ewilligers:
Added  "The user agent MAY display options's message to the user."

The following sentence also used the word "option" so I used a synonym.

level-2/index.html
```diff
@@ -195,8 +194,9 @@ <h4>
                 </li>
                 <li>Present the user with a choice of one or more <a>share
                 targets</a>, selected at the user agent's discretion. The user
-                MUST be given the option to cancel rather than choosing any of
-                the share targets. Wait for the user's choice.
+                agent MAY display options's message to the user. The user MUST
+                be given the latitude to cancel rather than choosing any of the
```
mgiuca: "latitude" seems like a strange word choice here (I guess you're trying to avoid the word "option"). Can you just use "choice"?

level-2/index.html
```diff
@@ -195,8 +194,9 @@ <h4>
                 </li>
                 <li>Present the user with a choice of one or more <a>share
                 targets</a>, selected at the user agent's discretion. The user
-                MUST be given the option to cancel rather than choosing any of
-                the share targets. Wait for the user's choice.
+                agent MAY display options's message to the user. The user MUST
```
mgiuca: Please format "options's message" as `<var>options</var>'s <a>message</a>` (same as "data's url" above).

ewilligers: done

mgiuca:
lgtm

raymeskhoury:
Just curious - would we implement this in Chrome? We tend to be very careful about allowing websites to display a message in trusted UI.

PR #86, opened 2019-01-09
Don't share if files is empty, with other fields absent

ewilligers:
navigator.share({}) already rejects with TypeError.
navigator.share({files: []}) now also rejects.

This came up during implementation
https://chromium-review.googlesource.com/c/chromium/src/+/1351340/2/third_party/blink/renderer/modules/webshare/navigator_share.cc#44

***
<a href="https://pr-preview.s3.amazonaws.com/ewilligers/web-share/pull/86.html" title="Last updated on Jan 18, 2019, 2:05 AM UTC (e6cf286)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/web-share/86/3510c0e...ewilligers:e6cf286.html" title="Last updated on Jan 18, 2019, 2:05 AM UTC (e6cf286)">Diff</a>

### DIFF
--- a/level-2/index.html
+++ b/level-2/index.html
@@ -148,10 +148,15 @@
             <li>Let <var>p</var> be <a data-cite=
             "!promises-guide#a-new-promise">a new promise</a>.
             </li>
-            <li>If none of <var>data</var>'s members <a>title</a>, <a>text</a>,
-            <a>url</a> or (if the implementation supports file sharing)
-            <a>files</a> are present, reject <var>p</var> with
-            <a>TypeError</a>, and abort these steps.
+            <li>If none of <var>data</var>'s members <a>title</a>, <a>text</a>
+            or <a>url</a> are present:
+              <ol>
+                <li>If <var>data</var>'s member <a>files</a> is not present or
+                is empty, or if the implementation does not support file
+                sharing, reject <var>p</var> with <a>TypeError</a>, and abort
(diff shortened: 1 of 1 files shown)


level-2/index.html
```diff
@@ -148,10 +148,15 @@ <h4>
...
-            <li>If none of <var>data</var>'s members <a>title</a>, <a>text</a>,
-            <a>url</a> or (if the implementation supports file sharing)
-            <a>files</a> are present, reject <var>p</var> with
-            <a>TypeError</a>, and abort these steps.
+            <li>If none of <var>data</var>'s members <a>title</a>, <a>text</a>
+            or <a>url</a> are present:
+              <ol>
+                <li>If <var>data</var>'s member <a>files</a> is not present or
+                is empty, or if the implementation does not support file
+                sharing, reject <var>p</var> with <a>TypeError</a>, and abort
```
raymeskhoury: I'm wondering whether it would be better to use a different type of error if file sharing isn't supported? This preserves the existing behavior though, so seems fine for now.

raymeskhoury: Also, should the title of this PR be "Don't share if files is empty..."?

ewilligers: Using TypeError is convenient here as canShare needs to return false, and canShare returns false whenever the share would fail with a TypeErrors.

PR #91, opened 2019-02-25
When blocking file types, reject with NotAllowedError.

### LINKED ISSUE #90: Specify behavior when a file type is rejected
Some user agents may wish to prevent certain file types, for example executable files, from being shared. This decision might be made on the basis of MIME type or file extension, and the platform.

For example, Blink identifies various file extensions as dangerous:  [download_file_types](https://cs.chromium.org/chromium/src/chrome/browser/resources/safe_browsing/download_file_types.asciipb).

Some user agents may wish to prevent certain sites from sharing, for example known phishing sites.

Should script be able to detect in advance that such share attempts will fail? If so, then `share` should throw a `TypeError` and `canShare` should return `false`. If not, then `share` should throw a `NotAllowedError` (like if there was no user gesture) or `AbortError` (like if the user cancelled or there were no suitable share targets) and `canShare` should return `true`.

ewilligers:
resolves #90

***
<a href="https://pr-preview.s3.amazonaws.com/ewilligers/web-share/pull/91.html" title="Last updated on Feb 25, 2019, 11:45 PM UTC (03b9002)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/web-share/91/d18ffab...ewilligers:03b9002.html" title="Last updated on Feb 25, 2019, 11:45 PM UTC (03b9002)">Diff</a>

### DIFF
--- a/level-2/index.html
+++ b/level-2/index.html
@@ -192,7 +192,8 @@
             <li>
               <a data-cite="!HTML#in-parallel">In parallel</a>:
               <ol>
-                <li>If there are no <a>share targets</a> available,
+                <li>If there are no <a>share targets</a> available, or a file
+                type is being blocked due to security considerations,
                 <a data-cite="!promises-guide#reject-promise">reject</a> <var>
                   p</var> with an "<a>AbortError</a>" <a>DOMException</a>, and
                   abort these steps.
--- a/level-2/index.html
+++ b/level-2/index.html
@@ -184,7 +184,8 @@
             </li>
             <li>If the method call was not <a data-cite=
             "!HTML#triggered-by-user-activation">triggered by user
-            activation</a>, <a data-cite=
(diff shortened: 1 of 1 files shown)


level-2/index.html
```diff
@@ -192,7 +192,8 @@ <h4>
             <li>
               <a data-cite="!HTML#in-parallel">In parallel</a>:
               <ol>
-                <li>If there are no <a>share targets</a> available,
+                <li>If there are no <a>share targets</a> available, or a file
+                type is being blocked due to security considerations,
```
raymeskhoury: nit: Does this fit better with the step above which checks for user activation?

ewilligers: Updated to reject with a NotAllowedError.

ewilligers:
We might also want to reject shares that are too large (large files and/or too many files).

PR #101, opened 2019-06-18
Demo now checks for https

### LINKED ISSUE #100: Demos don't check for https to make error messaging less obtuse
Heads up that to get to these demos, a search for "web share api test wicg github" on Google results in a link (wicg.github.io/web-share/demos/share-files.html) with `http:` scheme. The distinction between http and https is somewhat hidden on mobile still as compared to current desktop browsers.

Per the docs this API only works from `https:`. Because I didn't realize that I hadn't loaded https, it took a while to figure out if my browser needed Android-level app permissions for this to work or there was some other failure - more diagnostic if possible while enabling bleeding edge capability is always welcome.

Thanks for considering, and looking forward to using this API!

(Curiously the demo is the only link in these search results that's still on http by default)

Screen Shot 2019-06-14 at 9 56 52 PM

ewilligers:
If people arrive at the demo page via http://, we now
redirect to https instead of immediately reporting that
navigator.share is not supported.

resolves #100

### DIFF
--- a/demos/share-files.html
+++ b/demos/share-files.html
@@ -128,8 +128,16 @@
           testWebShareDelay);
 
       if (navigator.share === undefined) {
-        logError('Error: You need to use a browser that supports this draft ' +
-                 'proposal.');
+        if (window.location.protocol === 'http:') {
+          logError('navigator.share() is only available in secure contexts.');
+          setTimeout(() => {
+            window.location.replace(window.location.href.replace(/^http:/, 'https:'));
+          }, 4000);
+          return;
+        } else {
+          logError('Error: You need to use a browser that supports this draft ' +
+                   'proposal.');
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


demos/share-files.html
```diff
@@ -128,8 +128,16 @@ <h1>Web Share Test</h1>
           testWebShareDelay);
 
       if (navigator.share === undefined) {
-        logError('Error: You need to use a browser that supports this draft ' +
-                 'proposal.');
+        if (window.location.protocol === 'http:') {
+          logError('navigator.share() is only available in secure contexts.');
+          setTimeout(() => {
+            window.location.replace(window.location.href.replace(/^http:/, 'https:'));
+          }, 4000);
```
raymeskhoury: 4 seconds seems like a long time to wait for the redirect? Should we just make it happen straight away?

PR #102, opened 2019-06-19
Editorial: ReSpec fixes

marcoscaceres:


ewilligers:
level-2/index.html will also need to be updated.

marcoscaceres:
I wonder if we should just fold level 2 into level 1? At least, we should start discussing those additions as part of the base spec as we transition this over to the WebApps WG. If we get implementer commitment for the v2 set of things, then we probably don’t need multiple levels at this point.

PR #103, opened 2019-06-21
fix: algorithm aborts before promise is returned

marcoscaceres:


ewilligers:
We will need to make the same change to level-2/index.html

PR #111, opened 2019-07-03
fix: default share(data) to empty dictionary

### LINKED ISSUE #109: why "optional" in share(optional ShareData data)?
Given that the first thing we need to check is if `data` has some members, shouldn't the share() signature actually make `data` required? 

Alternatively, the method signature should default to the empty dictionary: 

```
[SecureContext] Promise<void> share(optional ShareData data = {});
```

However, I think we should just make `data` required.

marcoscaceres:
fixes #109

### DIFF
--- a/index.html
+++ b/index.html
@@ -90,7 +90,7 @@
         </h3>
         <pre class="idl">
           partial interface Navigator {
-            [SecureContext] Promise&lt;void&gt; share(optional ShareData data);
+            [SecureContext] Promise&lt;void&gt; share(optional ShareData data = {});
           };
         </pre>
         <div class="note">
--- a/index.html
+++ b/index.html
@@ -93,14 +93,6 @@
             [SecureContext] Promise&lt;void&gt; share(optional ShareData data = {});
           };
         </pre>
-        <div class="note">
-          The <var>data</var> argument is marked as "optional", but it is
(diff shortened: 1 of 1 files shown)


marcoscaceres:
We probably don't need the note anymore because now it's clear from the IDL that at least `{}` is always present. Now the algorithm always makes sense :)

marcoscaceres:
I can’t recall if the IDL changes are automatically picked up on WPT. Also, any idea if https://github.com/heycam/webidl/pull/750 is implemented in Blink yet... I’m guessing not, but we should probably track that.

ewilligers:
> any idea if heycam/webidl#750 is implemented in Blink yet?

No.
```
boolean canShare(optional ShareData data = {});
Promise<void> share(optional ShareData data = {});
```
gives
```
navigator_share.idl(12) : error: Unexpected "{" after "=".
navigator_share.idl(15) : error: Unexpected "{" after "=".
```

marcoscaceres:
WPT also hasn’t updated to the latest version of WebIDL2.js, which includes support for this. Should hopefully happen by end of the week. 

Do we have a Chrome bug we can track? Should we file one?

ewilligers:
Chrome bug for dictionaries with no required members:
https://bugs.chromium.org/p/chromium/issues/detail?id=948139

PR #113, opened 2019-07-08
Introduce [[sharePromise]] internal slot

### LINKED ISSUE #110: Clarifying behavior of multiple simultaneous shares
We probably want to clarify what happens when multiple concurrent shares are triggered by user activation. Consider: 

```JS
var b = document.createElement("button");
document.body.appendChild(b);
b.innerHTML = "hello!";
b.onclick = () => {
    var  p1 =  navigator.share({text: "i am 1", title: "i am 1"});
    var  p2 =  navigator.share({text: "i am 2", title: "i am 2"});
    p1.then(()=>console.log("1 resolved")).catch(err =>  console.error(1, err));
    p2.then(()=>console.log("2 resolved")).catch(err =>  console.error(2, err));
}
```

Should p1 win and p2 automatically gets rejected with an AbortError? or should p1 get rejected, and p2 wins? 

Safari currently shares p1, but rejects **both** with `AbortErrors` (even though the share is successful). That seems like a  bug. 

In any case, seems like if p1 should win and trying to make subsequent calls to share() should reject with an `InvalidStateError` (i.e., only allow one share at a time per document).

### LINKED ISSUE #112: Distinguishing between data errors and user abort
The spec says:

> Activate the chosen share target, convert data to a format suitable for ingestion into the target, and transmit the converted data to the target. If an error occurs starting the target or transmitting the data, reject p with an "AbortError" DOMException, and abort these steps.

We probably want a `DataError` error here, so a developer knows that it's not the user aborting, but something wrong with the data they are trying to pass.

marcoscaceres:
* fixes #110
* fixes #112 

 * [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/18486)

Implementation commitment:

 * [ ] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=201627)
 * [ ] [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1002337)
 * [x] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1312422)

For discussion... this prevents  #110 (multiple concurrent shares).

### DIFF
--- a/index.html
+++ b/index.html
@@ -106,6 +106,25 @@
         </div>
         <section>
           <h4>
+            Internal Slots
+          </h4>
+          <p>
+            This API adds the following internal slot to the {{Navigator}}
+            interface.
+          </p>
+          <dl>
+            <dt>
+              {{Promise}}? <dfn>[[\sharePromise]]</dfn>
+            </dt>
+            <dd>
+              The {{[[sharePromise]]}} is a promise that represents a user's
+              current intent to share some data with a <a>share target</a>. It
+              is initialized to `null`.
+            </dd>
+          </dl>
+        </section>
+        <section>
+          <h4>
             <dfn>share()</dfn> method
           </h4>
           <p>
@@ -113,6 +132,9 @@
             |data:ShareData|, run the following steps:
           </p>
           <ol class="algorithm">
+            <li>If {{[[sharePromise]]}} is not `null`, return <a>a promise
+            rejected with</a> {{InvalidStateError}}.
+            </li>
             <li>If none of |data|'s members {{ShareData/title}},
             {{ShareData/text}}, or {{ShareData/url}} are present, return <a>a
             promise rejected with</a> a {{TypeError}}.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -147,32 +170,61 @@ <h4>
             return <a>a promise rejected with</a> with a "{{NotAllowedError}}"
             {{DOMException}}.
             </li>
-            <li>Let <var>p</var> be <a>a new promise</a>.
+            <li>Set <a>[[\sharePromise]]</a> be <a>a new promise</a>.
```
mgiuca: s/be/to

index.html
```diff
@@ -112,6 +112,26 @@ <h3>
...
+          </p>
+          <dl>
+            <dt>
+              {{Promise}}? <dfn data-dfn-for=
+              "Navigator">[[\sharePromise]]</dfn>
+            </dt>
+            <dd>
+              The <a>[[\sharePromise]]</a> is a promise that represents a
+              user's current intent to share some data with a <a>share
+              target</a>. It is initialized to `null`.
```
mgiuca: Backticks won't have any effect, will they?

In my specs (at least WST), I just leave the word "null" unformatted. How do you format it?

mgiuca:
Oops forgot: You need to add these changes to Level 2 as well. L2 isn't a patch on L1, it's a "fork".

Or, just land this and I'll copy-paste it to L2 if you don't have the time (let me know).

index.html
marcoscaceres: ReSpec now formats it with `<code>null</code>`: 
Screenshot 2019-08-16 14 25 21

marcoscaceres:
Blocking myself on providing tests.

marcoscaceres:
If you don't mind, please update the L2 for me :)

mgiuca:
OK, after you're ready to merge (or after you merge), so that we don't have to keep syncing.

(Unfortunately, pr-preview didn't work because it only looks at level-2. I can't find a way for pr-preview to look at more than one document.)

marcoscaceres:
> Unfortunately, pr-preview didn't work because it only looks at level-2. I can't find a way for pr-preview to look at more than one document. 

Yeah, unfortunately, I don't think that's supported yet. @tobie can probably confirm?

marcoscaceres:
On Gecko side, part of https://bugzilla.mozilla.org/show_bug.cgi?id=1312422

marcoscaceres:
@mgiuca, will you handle filing bugs on the Chromium side?

tobie:
>> Unfortunately, pr-preview didn't work because it only looks at level-2. I can't find a way for pr-preview to look at more than one document.

> Yeah, unfortunately, I don't think that's supported yet. @tobie can probably confirm?

Yeah, that's not supported for now.

I'd be open to build the feature is someone was interested in sponsoring the work.

marcoscaceres:
@mgiuca or @ericwilligers, could you please review the associated test?

marcoscaceres:
@ericwilligers or @mgiuca, if you have a minute, could you please review web-platform-tests/wpt#18486?

marcoscaceres:
Filed Chrome bug https://bugs.chromium.org/p/chromium/issues/detail?id=1002337

marcoscaceres:
Filed WebKit bug https://bugs.webkit.org/show_bug.cgi?id=201627

marcoscaceres:
Fixed also in WebKit from yesterday https://bugs.webkit.org/show_bug.cgi?id=201627

Issue #108, opened 2019-07-01
Simplifying canShare()


marcoscaceres:
I wonder if we can simplify canShare() a bit... right now, canShare() conflates type checking behavior by intertwining the method call with share(). This is problematic because share() returns a promise, not TypeErrors. 

Over on Mozilla’s bugzilla, @ericwilligers mentioned that the intent of canShare() is to avoid having to do UA sniffing. As such, why don’t we this instead:

```
enum Shareable {
  "file”,
  "url”,
   // and so on...
}

partial interface Navigator {
   boolean canShare(sequence<Shareable> members);
}
```

That greatly simplifies things by deferring the support checking to the IDL layer. That causes the IDL layer to throw if passed some unsupported Shareable type. The result being that it either returns true or throws. 

If throwing is not palatable, then an alternative would be to take a sequence of DOMStrings, then return false if the names don’t match a supported member type.

raymeskhoury:
I like that rough idea because I can see that often sites would want to check this up front to determine whether to enable file sharing. 

Currently we also restrict the mime types and extensions that can be shared to a limited set. canShare could also be useful to determine if the site can share a particular type.

So perhaps canShare -> canShareFileTypes([array of file extensions/mime types]).

I'm interested in @mgiuca's thoughts here.

marcoscaceres:
This is somewhat reminiscent of the defunct `navigator.registerContentHandler()` ... obviously, this is different as we want to check mine type support... however, dealing with mime types gets extremely complicated very quickly; like, sharing video, but it's in some weird compression format. One share target may support a file type, but another might not (even if the UA supports the type). 

We might need to restrict the types... or maybe look at how other parts of the platform deals with this? For example, drag and drop must somehow deal with this somehow. And probably other parts of the platform too.... I'm looking at HTML Media Capture, and the interop story there also seems to be not great: https://caniuse.com/#feat=html-media-capture

Hmmm... 🤔

ewilligers:
>  canShare() conflates type checking behavior by intertwining the method call with share(). This is problematic because share() returns a promise, not TypeErrors.

We simply wanted to avoid specifying the same list of conditions in two different places, to avoid the risk of them ever getting out of sync. I see the following as equivalent:-

1. Listing the conditions in share(), and saying canShare() returns false iff share() would reject with TypeError.

1. Listing the conditions in canShare(), and saying share() rejects with TypeError iff canShare would return false.

[...]

PR #118, opened 2019-08-15
Do 'triggered by user activation' check earlier

marcoscaceres:

Implementation commitment:

### DIFF
--- a/index.html
+++ b/index.html
@@ -113,6 +113,10 @@
             <var>data</var>, run the following steps:
           </p>
           <ol data-link-for="ShareData">
+            <li>If the method call was not <a>triggered by user activation</a>,
+            return <a>a promise rejected with</a> with a {{"NotAllowedError"}}
+            {{DOMException}}.
+            </li>
             <li>If none of <var>data</var>'s members <a>title</a>, <a>text</a>,
             or <a>url</a> are present, return <a>a promise rejected with</a> a
             {{TypeError}}.
@@ -136,10 +140,6 @@
                 serializer</a> on <var>url</var>.
                 </li>
               </ol>
-            </li>
-            <li>If the method call was not <a>triggered by user activation</a>,
-            return <a>a promise rejected with</a> with a {{"NotAllowedError"}}
-            {{DOMException}}.
             </li>
             <li>Let <var>p</var> be <a>a new promise</a>.
             </li>


marcoscaceres:
Would prefer to do this check earlier in Gecko if possible.

marcoscaceres:
Safari does the check later, so Gecko will follow Chrome and Safari.

mgiuca:
IIRC the reason why this check comes after all the TypeErrors is a _little_ dubious but deliberate: it makes it possible to write automated error tests that check for TypeErrors in the [WPT suite](https://github.com/web-platform-tests/wpt/tree/master/web-share). All tests that expect us to go past the user activation check have to be manual tests, because they will all fail if automated.

If the user activation check is moved up ahead, all the automated tests will fail with NotAllowedError.

marcoscaceres:
> IIRC the reason why this check comes after all the TypeErrors is a little dubious but deliberate: it makes it possible to write automated error tests that check for TypeErrors in the WPT suite. All tests that expect us to go past the user activation check have to be manual tests, because they will all fail if automated.

This was solved on WTP via Web Driver and `.bless()` (or just using `.click()`, which fires trusted events when run under automation).

marcoscaceres:
This came up again during review on the Gecko side... given that we can deal with this correctly on WTP, @mgiuca would you be willing to change the Blink implementation?

ewilligers:
Are you also proposing that `canShare` (if merged) would also require user activation? If not, what is the benefit of moving the check up?

I don't mind changing the order in the Blink implementation if that is the consensus. Any feedback from Safari?

marcoscaceres:
> Are you also proposing that canShare (if merged) would also require user activation? 

No. But in the related issue (I think #5), I discussed how to separate the concerns between sharing from canShare.

> If not, what is the benefit of moving the check up?

It's just convention in Gecko code and other APIs to do the "triggered by user activation" checks first - before doing all the data checks. So the benefit would be consistency with other APIs (and less Gecko reviewers telling me "move this up" 😂).

mgiuca:
It doesn't make sense for `canShare` to require user activation.

I think the reason Eric raises this is that `canShare` is currently [defined in the L2 spec as](https://w3c.github.io/web-share/level-2/#canshare-method):

> `canShare(data)` MUST return true unless `share(data)` would reject with TypeError, in which case it MUST return false.

That definition wouldn't work if we moved the user activation check up. But I think Marcos also objected to that definition (somewhere else) and that we should define it properly by abstracting out the type checking. Which I can get behind.

That aside, it also seems logical to me that the method would do input validation (i.e., checks that are pure mathematical function of the input) before doing extrinsic checks that rely on the state-of-the-world (i.e., checks about user activation).

> It's just convention in Gecko code and other APIs to do the "triggered by user activation" checks first - before doing all the data checks.

If that's the case, is it also a convention in other web APIs (specs)? Can you point to other web specs that do user activation testing before input validation? If so, I'd be happy to change it.

marcoscaceres:
> That definition wouldn't work if we moved the user activation check up. But I think Marcos also objected to that definition (somewhere else) 

Indeed. I think it was #108. 

> and that we should define it properly by abstracting out the type checking. Which I can get behind.

That would be great. It means doing less gymnastics in our implementation. 

> That aside, it also seems logical to me that the method would do input validation (i.e., checks that are pure mathematical function of the input) before doing extrinsic checks that rely on the state-of-the-world (i.e., checks about user activation).

I don't disagree... we kinda do this already:  

 1. binding layer does type checks,  which, if you squint just right, would be the "pure mathematical function of the input". 

But then we diverge on: 

 1. Security/input checks, including triggered by user activation, possibly feature policy checks, etc.. 
 1. Do actual data mutations (e.g., resolve URL).
 1. IPC... and so on.... 

> If that's the case, is it also a convention in other web APIs (specs)? Can you point to other web specs that do user activation testing before input validation? If so, I'd be happy to change it.

Some examples:
 
* (biased, I wrote it) https://www.w3.org/TR/payment-request/#show-method
* (biased, I wrote it) https://w3c.github.io/screen-orientation/#apply-an-orientation-lock
* https://html.spec.whatwg.org/#file-upload-state-(type=file)

Somewhat in the middle:

 * ("allowed to request fullscreen", relies on "triggered by use activation") https://fullscreen.spec.whatwg.org/#dom-element-requestfullscreen

mgiuca:
Discussed offline; we are OK to keep it for now (those examples aren't particularly strong since they don't go TypeErrors after the activation check).

PR #131, opened 2019-10-20
Clarify that UAs can concatenate fields like on Android.

### LINKED ISSUE #130: Clarify how URL should be used on platforms without dedicated URL field
Similar to [Web Share Target's note](https://wicg.github.io/web-share-target/#obtaining-a-sharedata), the spec should clarify how to handle the URL property on platforms without a dedicated URL field (like Android). 

The desired behaviour is probably to put text and URL in the same field, since Web Share Target expects that.

mgiuca:
Moved the non-normative explanation out of the Examples section into
the Share targets section, and also addded normative directive that user
agents may discard *or combine* members (emphasis on the new text).

Closes #130.

### DIFF
--- a/index.html
+++ b/index.html
@@ -305,7 +305,7 @@
         of the payload is at the discretion of the share target.
       </p>
       <div class="note">
-        Mapping the {{ShareData}} to the share target (or operating system)'s
+        Mapping the {{ShareData}} to the share target's (or operating system's)
         native format can be tricky as some platforms will not have an
         equivalent set of members. For example, if the target has a "text"
         member but not a "URL" member (as is the case on Android), one solution
--- a/index.html
+++ b/index.html
@@ -301,9 +301,18 @@
         concepts exposed in {{ShareData}}. To <dfn>convert data to a format
         suitable for ingestion into the target</dfn>, the user agent SHOULD map
         the members of {{ShareData}} onto equivalent concepts in the target. It
-        MAY discard members if necessary. The meaning of each member of the
-        payload is at the discretion of the share target.
+        MAY discard or combine members if necessary. The meaning of each member
+        of the payload is at the discretion of the share target.
       </p>
+      <div class="note">
+        Mapping the {{ShareData}} to the share target (or operating system)'s
+        native format can be tricky as some platforms will not have an
+        equivalent set of members. For example, if the target has a "text"
+        member but not a "URL" member (as is the case on Android), one solution
+        is to concatenate both the {{ShareData/text}} and {{ShareData/url}}
+        members of {{ShareData}} and pass the result in the "text" member of
+        the target.
+      </div>
       <p>
         Each share target MAY be made conditionally available depending on the
         {{ShareData}} payload delivered to the {{Navigator/share()}} method.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -301,9 +301,18 @@ <h2>
         concepts exposed in {{ShareData}}. To <dfn>convert data to a format
         suitable for ingestion into the target</dfn>, the user agent SHOULD map
         the members of {{ShareData}} onto equivalent concepts in the target. It
-        MAY discard members if necessary. The meaning of each member of the
-        payload is at the discretion of the share target.
+        MAY discard or combine members if necessary. The meaning of each member
+        of the payload is at the discretion of the share target.
       </p>
+      <div class="note">
+        Mapping the {{ShareData}} to the share target (or operating system)'s
```
ewilligers: share target's (or operating system's)

marcoscaceres:
Another odd situation is where OS requires a field, like `title` is required on Windows. Right now, we are hacking around it with a single “ “... but it’s not great.

mgiuca:
> Another odd situation is where OS requires a field, like `title` is required on Windows. Right now, we are hacking around it with a single “ “... but it’s not great.

What does Windows do with the title? (It seems strange that it's required, since often you just share a single URL by itself.) Is it possible that "title" has a different meaning in this context? (For instance, is it the title of the dialog box, as opposed to the title of the content being shared?)

marcoscaceres:
@saschanaz, any chance you could provide us with a screenshot? I think it wants to display it in the title bar of the share UI itself.

mgiuca:
> @saschanaz, any chance you could provide us with a screenshot? I think it wants to display it in the title bar of the share UI itself.

Ah if that's the case then that's quite a different use case, and you shouldn't be using the `title` field (if supplied) for that. You should probably just have a fixed string like "Share via" or "Share from `origin`" or something. We could consider having a separate field like `dialogTitle` which the UA can optionally use to set the dialog title, but we must beware of spoofing.

saschanaz:
>I think it wants to display it in the title bar of the share UI itself.

Yes AND it also uses it as a title of content:

image

image

marcoscaceres:
Oooh.... yeah, ok... we might need to clarify some things in the spec. We got a single field being used for multiple things.

saschanaz:
>since often you just share a single URL by itself.

The current Firefox global share menu sets the title as `document.title` when a user want to share the current URL, because Windows 🤷‍♀️

ewilligers:
> we must beware of spoofing.

There was an earlier request https://github.com/w3c/web-share/pull/85 for authors to specify some text that would appear with the share sheet. We reverted it due to spoofing concerns.

mgiuca:
> Yes AND it also uses it as a title of content:

Oh OK, in that case I think using `title` here is appropriate. Basically the title of the dialog is "Share" with a sub-heading that says "You are going to share an article titled "Jabberwocky"".

I think it's fine to not clarify it in the spec since this is a Windows-UI-specific usage of the "title" field. It is frustrating though, that they require it. I wonder if we could ask the Windows team to make it optional? @aarongustafson ?

marcoscaceres:
> It is frustrating though, that they require it. 

Agree. It would be nice if it wasn't and Windows just applied as default title.

PR #133, opened 2019-11-06
Add the ability to share files

marcoscaceres:
Add a `files` member to ShareData dictionary.

@ericwilligers, I tried to add in what is in level 2 to kick off discussion. 

 * [X] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/23423)

Implementation commitment:

 * [X] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=207491)
 * [X] Chromium - Already implemented.
 * [ ] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1635700)

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,7 +32,7 @@
       };
     </script>
   </head>
-  <body>
+  <body data-cite="FILEAPI">
     <section id="abstract">
       <p>
         This specification defines an API for sharing text, links and other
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -238,6 +248,7 @@ <h3>
         </h3>
         <pre class="idl">
           dictionary ShareData {
+            FrozenArray&lt;File&gt; files;
```
marcoscaceres: I think this should be a `sequence`, no? We wouldn't expect a developer to pass in a frozen array (i.e., `{files: Object.freeze(filesArray)}`). 
```suggestion
            sequence&lt;File&gt; files;
```

marcoscaceres:
@hober, during TPAC you mentioned that this may be of interest to the WebKit community. Would you be the right person to help us with reviewing/designing this extension to web share?

othermaciej:
There's code implementing file sharing now in WebKit trunk, see: https://bugs.webkit.org/show_bug.cgi?id=207491

marcoscaceres:
@othermaciej, that's fantastic new! thanks for the update. 

Were you folks ok with the way `.canShare()` works? I'd made an alternative proposal in #108? Happy to close it that if WebKit will follow Chrome's behavior tho.

othermaciej:
We've currently implemented more or less what is in the spec. However, it's not too late to change what we have before it ships in a non-preview release.

I personally think the change proposed in issue #108 would be an improvement. I think `canShare` at one point checked for support of sharing specific file types, but now effectively just checks whether file sharing is supported. It seems unwieldy to construct a full ShareData just to check that. One downside is that a method with the same name but different parameters would be maximally likely to cause a compat issue if any content has adopted this AP. I'll comment to that effect in the issue.

marcoscaceres:
Sent a PR to WPT so the corresponding test is [no longer be tentative](https://github.com/web-platform-tests/wpt/pull/23423). 

Filed [Gecko bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1635700). I'll see if I, or someone on Mozilla's DOM Team, can find time to implement it.

marcoscaceres:
@ericwilligers, @mgiuca, this is probably ready to land I think.

PR #134, opened 2019-11-18
chore: inform people that l2 is folded into main spec

### LINKED ISSUE #107: Drop the levels/versioning
For when we move to Web Apps WG, we should consider dropping the spec Levels... instead, we should make pull requests for the level 2 features and try to get other browser vendors behind it. It seems like we (Mozilla) might be willing to support `.canShare()` and sharing `File`s out of the box, so that would give us two implementations (bar review of the actual features, that we are yet to do)... we would then need to sell the Safari folks on supporting those features 🤞.

marcoscaceres:
Closes #107 

The level 2 spec is starting to show up in places, so we should merge this sooner rather than later to minimize the damage.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 22, 2020, 10:51 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fweb-share%2F034c2618a3e8d26aa0c15a71915b0b06db87ff84%2Flevel-2%2Findex.html%3FisPreview%3Dtrue)

```
[33m🕵️‍♀️  That doesn't seem to be a ReSpec document. Please check manually:[39m [36m[Function: url][39m
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/web-share%23134.)._
</details>


mgiuca:
It isn't yet folded into the spec, right? So this is just deleting content that isn't available anywhere else (then we'll have to fish it out of revision control when we fold it into the spec). I think we should hold off on this.

> to minimize the damage

What damage? If we eventually make the l2 spec just tell people to look at the l1 spec (i.e., land this change), then it doesn't matter if people are linking to the l2 spec for now.

marcoscaceres:
> It isn't yet folded into the spec, right? 

Correct, we need to land #133. 

> What damage?

Mostly broken links. We have for instance: 
https://wiki.developer.mozilla.org/en-US/docs/Web/API/Navigator/canShare

(we don't have consensus on `canShare()` or the shape of that API - #108)

Which unfortunately was prematurely documented. Similarly talks about L2:
https://developers.google.com/web/updates/2019/05/web-share-files

> then it doesn't matter if people are linking to the l2 spec for now.

Depends. The L2 will be out of sync with the published version and end up causing confusion (e.g., in the case of MDN).

marcoscaceres:
@mgiuca, I'd really like for us to get this merged as `canShare()` keeps coming up and it's not part of the main spec. We can then finally get consensus on what `canShare()` will look like and how it will behave and land that in the main spec.

marcoscaceres:
Merging as https://github.com/w3c/web-share/pull/177 is now in place.

sideshowbarker:
2bbde9c broke a link in MDN to the Level 2 spec — in the Specifications table here:

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/canShare#Specifications

…which links to https://w3c.github.io/web-share/level-2/#canshare-method

And since #177 hasn’t been merged yet, we don’t have any other working URL that we could change that link to.

marcoscaceres:
@sideshowbarker, `canShare()` might not make it into the spec. It's still tentative. Mozilla will probably object to it being included for the reasons outlined in #108.  See #184 as the proposed alternative.

marcoscaceres:
(i.e., canShare() was never part of an official spec, it was just an incubation... it should not have been documented on MDN as anything but experimental... why I talked about merging this back November 2019, "to minimize the damage").

sideshowbarker:
> `canShare()` might not make it into the spec. It's still tentative (and not fit for purpose). Mozilla will probably object to it being included for the reasons outlined in #108. See #184 as the proposed alternative.
>
> (i.e., canShare() was never part of an official spec, it was just an incubation... it should not have been documented on MDN as anything but experimental... why I talked about merging this back November 2019, "to minimize the damage").

OK, thanks — given all that, I’ve raised https://github.com/mdn/browser-compat-data/pull/6844 to mark it as experimental and nonstandard in BCD, and in https://wiki.developer.mozilla.org/en-US/docs/Web/API/Navigator/canShare$compare?to=1647398&from=1642438, I marked the MDN article similarly.

PR #135, opened 2019-12-13
Editorial: Use new HTML activation model

marcoscaceres:
This is refactor to use HTML's new activation model. The api currently uses transient activation.

Details: https://github.com/whatwg/html/issues/5129 

In a new issue, we need to discuss if calling share should "consume" the activation.

Doing so might have the nice effect that we can get rid of the internal slot we added to protect against being able to call `.share()` twice during the same activation phase.

### DIFF
--- a/index.html
+++ b/index.html
@@ -156,9 +156,9 @@
                 </li>
               </ol>
             </li>
-            <li>If the method call was not <a>triggered by user activation</a>,
-            return <a>a promise rejected with</a> with a {{"NotAllowedError"}}
-            {{DOMException}}.
+            <li>If the [=relevant global object=] of [=this=] does not have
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@mgiuca, @ericwilligers, this is somewhat high-priority for getting the document published this week as FPWD. Happy to discuss or clarify anything - if we need to jump on quick hangout, also happy to do that.

marcoscaceres:
As we’ve been approved for publication by the W3C, and this doesn’t change make a substantive change, I’m going to merge this as it will otherwise the broken “triggered by user activation” link will block us from publishing until next year (I’m away until February).

marcoscaceres:
Thanks for checking it, @ericwilligers.

PR #137, opened 2019-12-13
BREAKING CHANGE: make share() consume user activation

### LINKED ISSUE #136: Should calling share() "consume" the activation?
HTML has a new use activation model that helps deal with situations like:

```JS
button.onclick = () => {
  const pShare = navigator.share(stuff);
  // should this reject? 
  const pFS = document.body.requestFullscreen();
}
```

Given the somewhat modal nature of the share sheet/context-popup menu (Safari desktop), it might make sense to consume the activation?  

Doing so might have the nice effect that we can get rid of the internal slot we added to protect against being able to call .share() twice during the same activation phase.

This would be a small breaking change for the currently implemented behavior.

marcoscaceres:
Closes #136 

Implementation commitment:

 * [ ] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1083037)
 * [ ] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1643205)

### DIFF
--- a/index.html
+++ b/index.html
@@ -443,11 +443,6 @@
         guard against this, but implementors will want to be aware that it is a
         possibility.
         </li>
-        <li>To prevent multiple share UIs from being displayed simultaneously,
-        as well as to mitigate some other behavior that could confuse end
-        users, calling {{Navigator/share()}} [=consume user activation|consumes
-        the user activation=] of the relevant {{Window}}.
-        </li>
       </ul>
     </section>
     <section class="appendix informative">
--- a/index.html
+++ b/index.html
@@ -170,10 +170,9 @@
                 </li>
               </ol>
             </li>
-            <li>If the [=relevant global object=] of [=this=] does not have
-            [=transient activation=], or a file type is being blocked due to
-            security considerations, return <a>a promise rejected with</a> with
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Blocked on https://github.com/w3c/web-share/pull/135 (it includes 135, so that should be merged first).

marcoscaceres:
This is no longer blocked.

marcoscaceres:
This is now unblocked... @mgiuca and @ericwilligers, what are your thoughts on supporting this? Seems fairly logical that we would consume the activation here, given that on most platforms this is a modal dialog.

marcoscaceres:
Filed [Gecko bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1643205) to implement this.

marcoscaceres:
Ok great... about the test, I see some testing suggestions. Should we try to come up with a WPT test together? Like we could check against some other API we all implement that also consumes the activation.

marcoscaceres:
Ok, I'm about ready to land this in Gecko including updated Web Platform Tests.  https://phabricator.services.mozilla.com/D79640

I'd expect them to show up in WPT in about 72 hours.

PR #150, opened 2020-04-29
Initial take on Security/Privacy self-review

### LINKED ISSUE #149: Complete privacy/security self-review
We need to complete a privacy and security self-review before we can request an official review by the Privacy IG and/or TAG.

There is a [self-review questionnaire](https://w3ctag.github.io/security-questionnaire/)

Related issue w3c/webappswg#27

inexorabletash:
Closes #149

I went through the S&P self review questionnaire and did my best to answer the questions.

### DIFF
--- a/docs/security-privacy-self-review.md
+++ b/docs/security-privacy-self-review.md
@@ -1,0 +1,80 @@
+
+
+Responses to the [Self-Review Questionnaire: Security and Privacy](https://w3ctag.github.io/security-questionnaire/) for the [Web Share API](https://w3c.github.io/web-share/)
+
+
+## 2.1 What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
+
+Web Share enables data to be sent from web sites to native applications. It explicitly does not allow the web site to learn which share targets are available.
+
+The API returns a rejected promise if no share targets are available, but also returns a rejected promise of the same type if the user cancels the share action. This intentionally does not reveal the state where no share targets are available.
+
+## 2.2 Is this specification exposing the minimum amount of information necessary to power the feature?
+
+The API returns three states: no sharing was attempted (via `AbortError`), sharing was attempted but failed (via `DataError`) or sharing was successful (no error). All three states seem necessary to inform the user of the result of the action.
+
+Implementors will want to carefully consider what information is revealed in the error message when `share()` is rejected. Even distinguishing between the case where no targets are available and user cancellation could reveal information about which apps are installed on the user's device.
+
+## 2.3 How does this specification deal with personal information or personally-identifiable information or information derived thereof?
+
+Any information shared was already exposed to the web site; no further personal information is revealed by the API to the web site.
+
+The receiving share target selected by the user can be given arbitrary information by the web site via the `ShareData`'s `title`, `text` and/or `url` properties. In most expected user interfaces, the user is not made aware of what data is being shared by the web site with the target. This could lead to a web site in posession of personal information sharing it with a target without the user's knowledge. One mitigation would be to have the user interface allow inspection of the data.
+
+## 2.4 How does this specification deal with sensitive information?
+
+As in the previous question, the specification does not limit what data is shared, and the data could include sensitive information already revealed to the web site. However, the user is in control of which, if any, target application will receive the information.
+
+## 2.5 Does this specification introduce new state for an origin that persists across browsing sessions?
+
+No.
+
+## 2.6 What information from the underlying platform, e.g. configuration data, is exposed by this specification to an origin?
+
+The presence of the API may reveal characteristics of the user's device beyond the user agent. For example, the user agent may only expose the `navigator.share()` API on "mobile" devices, revealing the type of device the user agent is executing on.
+
+## 2.7 Does this specification allow an origin access to sensors on a user’s device
+
+No.
+
+## 2.8 What data does this specification expose to an origin? Please also document what data is identical to data exposed by other features, in the same or different contexts.
+
+The data exposed to an origin is the presence of the API itself, and (per above) the results of interacting with the API (error or success).
+
+## 2.9 Does this specification enable new script execution/loading mechanisms?
+
+No.
+
+## 2.10 Does this specification allow an origin to access other devices?
+
+No.
+
+## 2.11 Does this specification allow an origin some measure of control over a user agent’s native UI?
+
+The API provides for showing a target selection user interface, for example as a modal dialog. The presence and presentation of the interface remains entirely under the control of the user agent.
+
+## 2.12 What temporary identifiers might this this specification create or expose to the web?
+
+None.
+
+## 2.13 How does this specification distinguish between behavior in first-party and third-party contexts?
+
+The contexts are not distingushed.
+
+The use of the feature is gated on [transient activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation) and thus requires user interaction (or the equivalent) within a third-party context to be used.
+
+## 2.14 How does this specification work in the context of a user agent’s Private Browsing or "incognito" mode?
+
+Use of `share()` from a private browsing mode might leak private data to a third-party application that does not respect the user's privacy setting. User agents could present additional warnings or disable the feature entirely when in a private browsing mode, but this is not mandated as the chooser UI could be considered sufficient warning.
+
+## 2.15 Does this specification have a "Security Considerations" and "Privacy Considerations" section?
+
+Yes.
+
+## 2.16 Does this specification allow downgrading default security characteristics?
+
+No.
+
+## 2.17 What should this questionnaire have asked?
+
+Exfiltration of data already exposed from the web site to contexts beyond the control of the user agent - in this case, native applications - could be called out, with potential risks and mitigations enumerated.
--- a/docs/security-privacy-self-review.md
+++ b/docs/security-privacy-self-review.md
@@ -5,7 +5,7 @@ Responses to the [Self-Review Questionnaire: Security and Privacy](https://w3cta
 
 ## 2.1 What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
 
-Web Share enables data to be sent from web sites to native applications. It explicitly does not allow the web site to learn which share targets are available.
+Web Share enables data to be sent from web sites to native applications or potentially other installed web applications. It explicitly does not allow the web site to learn which share targets are available.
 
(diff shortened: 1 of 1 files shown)


inexorabletash:
Note that I was looking at the non-level2 spec. It may be necessary to revisit...

inexorabletash:
Updated for level-2. I think we should probably add some things into the spec's privacy considerations section about how `canShare()` could be used for fingerprinting, and how user agents should mitigate that.

ewilligers:
Many thanks for filling out the survey.

`canShare()` does not reveal if there are apps installed that support given file types, or even if shares of a file type would be blocked for security reasons. It is more of a feature detection API, for example it reveals if the user agent supports sharing files. It should less useful for fingerprinting than `CSS.supports()`.

docs/security-privacy-self-review.md
```diff
@@ -0,0 +1,85 @@
+
+
+Responses to the [Self-Review Questionnaire: Security and Privacy](https://w3ctag.github.io/security-questionnaire/) for the [Web Share API](https://w3c.github.io/web-share/)
+
+
+## 2.1 What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
+
+Web Share enables data to be sent from web sites to native applications. It explicitly does not allow the web site to learn which share targets are available.
```
marcoscaceres, line 8:
```suggestion
Web Share enables data to be sent from web sites to native applications or potentially other installed web applications. It explicitly does not allow the web site to learn which share targets are available.
```

docs/security-privacy-self-review.md
```diff
@@ -0,0 +1,85 @@
...
+
+Implementors will want to carefully consider what information is revealed in the error message when `share()` is rejected. Even distinguishing between the case where no targets are available and user cancellation could reveal information about which apps are installed on the user's device. The previous version of the specification distinguished when sharing failed via `DataError` instead, which may have exposed unnecessary information.
+
+## 2.3 How does this specification deal with personal information or personally-identifiable information or information derived thereof?
+
+Any information shared was already exposed to the web site; no further personal information is revealed by the API to the web site.
+
+The receiving share target selected by the user can be given arbitrary information by the web site via the `ShareData`'s properties. In most expected user interfaces, the user is not made aware of what data is being shared by the web site with the target. This could lead to a web site in posession of personal information sharing it with a target without the user's knowledge. One mitigation would be to have the user interface allow inspection of the data.
```
marcoscaceres, line 24: On MacOS, there is always an intermediary UI before the sharing happens where the user can inspect the data. This is not to say that this advisory is incorrect (it's not) - but maybe we can drop "most".

Screenshot 2020-04-30 11 14 31

```suggestion
The receiving share target selected by the user can be given arbitrary information by the web site via the `ShareData`'s properties. In some expected user interfaces, the user is not made aware of what data is being shared by the web site with the target. This could lead to a web site in posession of personal information sharing it with a target without the user's knowledge. One mitigation would be to have the user interface allow inspection of the data.
```

docs/security-privacy-self-review.md
```diff
@@ -0,0 +1,85 @@
...
+
+## 2.5 Does this specification introduce new state for an origin that persists across browsing sessions?
+
+No.
+
+## 2.6 What information from the underlying platform, e.g. configuration data, is exposed by this specification to an origin?
+
+The presence of the API may reveal characteristics of the user's device beyond the user agent. For example, the user agent may only expose the `navigator.share()` API on "mobile" devices, revealing the type of device the user agent is executing on.
+
+Revealing content types supported via `files` should be done carefully, as `canShare()` does not require
```
marcoscaceres, line 38: Note that we have a proposal to mitigate this: #108. That is, canShare() relates to just checking support for what is supported by the API, not by the OS.

docs/security-privacy-self-review.md
```diff
@@ -0,0 +1,85 @@
...
+[transient activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation). For example, if the `canShare()` API response was conditional on what native applications were installed on the device, this could be used to fingerprint the user.
+
+## 2.7 Does this specification allow an origin access to sensors on a user’s device
+
+No.
+
+## 2.8 What data does this specification expose to an origin? Please also document what data is identical to data exposed by other features, in the same or different contexts.
+
+The data exposed to an origin is the presence of the API itself, whether certain content types are supported, and the results of interacting with the API (error or success).
```
marcoscaceres, line 47: Again, #108 might mitigate "whether certain content types are supported" to a degree. It would be good reinforce that a site can't distinguish between the user cancelled and no share target supports that content type.

docs/security-privacy-self-review.md
```diff
@@ -0,0 +1,85 @@
...
+
+No.
+
+## 2.10 Does this specification allow an origin to access other devices?
+
+No.
+
+## 2.11 Does this specification allow an origin some measure of control over a user agent’s native UI?
+
+The API provides for showing a target selection user interface, for example as a modal dialog. The presence and presentation of the interface remains entirely under the control of the user agent.
```
marcoscaceres, line 59:
```suggestion
The API provides for showing a target selection user interface, for example as a modal dialog on some platforms, or a context menu in other platforms. The presence and presentation of the interface remains entirely under the control of the user agent.
```

docs/security-privacy-self-review.md
```diff
@@ -0,0 +1,85 @@
...
+## 2.12 What temporary identifiers might this this specification create or expose to the web?
+
+None.
+
+## 2.13 How does this specification distinguish between behavior in first-party and third-party contexts?
+
+First-party and third-party contexts are not distingushed.
+
+The use of the `share()` API is gated on [transient activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation) and thus requires user interaction (or the equivalent) within a third-party context to be used.
+
```
marcoscaceres, line 70: @mgiuca, I can't recall, but do is the API disabled by default in third-party contexts (i.e., only enabled by feature policy)?

marcoscaceres:
Again noting that we should make a move to get rid of level 2 as per #107 (and have a single spec). We can avoid the whole l1/l2 confusion.

docs/security-privacy-self-review.md
mgiuca: I don't think there's any such restriction (either in the spec, or in Chrome's implementation).

Not sure what "requires user interaction (or the equivalent) within a third-party context to be used" is getting at. It always requires user interaction, why call out "within a third-party context" here?

mgiuca:
> Again noting that we should make a move to get rid of level 2 as per #107 (and have a single spec). We can avoid the whole l1/l2 confusion.

That's only feasible if we promote all of the L2 features into the spec. I don't want them taken out of the L2 spec and moved into PRs that linger for a long time while they are debated.

marcoscaceres:
>  I don't want them taken out of the L2 spec and moved into PRs that linger for a long time while they are debated.

Understood. But we should then make it more clear in the L2 document that what is only supported by a single engine (related https://github.com/WICG/admin/issues/102 -- cc @hober, as this has come up again, and we are trying to figure out a solution).

marcoscaceres:
Should we move this into the spec itself?

docs/security-privacy-self-review.md
inexorabletash: It is phrased this way to make it clear that even though third-party iframes can use `share()` - which might be surprising - the user still must have interacted with the frame. So an embedded ad frame is not going to be able to annoy the user, unless the user has interacted with the frame itself.

inexorabletash: (I admit to being surprised that we don't require feature policy for 3P contexts when I read the spec.)

inexorabletash:
> Many thanks for filling out the survey.

No problem! It's a chore... you should see the I18N one though. Yikes.

> `canShare()` does not reveal if there are apps installed that support given file types, or even if shares of a file type would be blocked for security reasons. It is more of a feature detection API, for example it reveals if the user agent supports sharing files.

This is pretty subtle and we should call it out explicitly both in the normative text near the definition, and in the Security and Privacy section. Implementations should be warned against making the test more precise for these reasons, and developers (if you have dev-focused text like domintro blocks) should be told that it's not that granular of a test.

docs/security-privacy-self-review.md
inexorabletash: Great! I'll reword, given the assumption that this lands.

inexorabletash:
> Should we move this into the spec itself?

My guidance would be to (1) keep this as a separate doc in the repo, for quick reference during horizontal review, but (2) look for any juicy bits that aren't covered in the spec's existing S&P section and copy them in.

The questionnaire responses end up being repetitive so I don't think just copy/pasting the whole thing into the spec will result in something that's very readable.

docs/security-privacy-self-review.md
inexorabletash: Thanks - updated!

docs/security-privacy-self-review.md
marcoscaceres: Agree. I think we should treat this as a bug.

marcoscaceres:
@inexorabletash sounds like a plan!

docs/security-privacy-self-review.md
```diff
@@ -0,0 +1,87 @@
...
+The API provides for showing a target selection user interface, for example as a modal dialog on some platforms, or a context menu in other platforms. The presence and presentation of the interface remains entirely under the control of the user agent.
+
+## 2.12 What temporary identifiers might this this specification create or expose to the web?
+
+None.
+
+## 2.13 How does this specification distinguish between behavior in first-party and third-party contexts?
+
+First-party and third-party contexts are not distingushed.
```
marcoscaceres, line 69:
```suggestion
First-party and third-party contexts are not distinguished. However, the working group is exploring limiting this API to first-party contexts and only allow the API to be exposed to third party contexts via feature policy. See [issue 51](https://github.com/w3c/web-share/issues/151) 
```

inexorabletash:
So this doc now covers level-2, which is mostly a superset of level-1 (there's a difference in error handling w/r/t `DataError` in L1 vs. `AbortError` in L2). 

Do we think that matters for the purposes of #149 ? Should there be two questionnaires? Should there be one but call out level-specific items? Apologies for barging in trying to help but not being cognizant of the nuances around the spec levels before starting...

docs/security-privacy-self-review.md
marcoscaceres: resolved based on us having filed https://github.com/w3c/web-share/issues/151

inexorabletash:
I can't merge in this repo, so if everyone is happy with this can someone hit the button?

PR #153, opened 2020-05-15
chore: Clarify reason for choosing USVString

### LINKED ISSUE #152: Note about choice of USVString imprecise
WebShare
https://w3c.github.io/web-share/#sharedata-dictionary

> These members are USVString (as opposed to DOMString) because they are not allowed to contain invalid UTF-16 surrogates. This means the user agent is free to re-encode them in any Unicode encoding (e.g., UTF-8). 

The above "Note" about the choice of USVString over DOMString is imprecise. I18N would suggest the following wording instead:

> These members are USVString (as opposed to DOMString) because they are not allowed to contain surrogate code points. Among other things, this means that the user agent can serialize them into any Unicode encoding, such as UTF-8, without change or loss of data or the generation of replacement characters.

ewilligers:
We use USVString members as they are not allowed to contain
surrogate code points.

closes #152

### DIFF
--- a/index.html
+++ b/index.html
@@ -293,10 +293,11 @@
         </dl>
         <div class="note">
           These members are {{USVString}} (as opposed to {{DOMString}}) because
-          they are not allowed to contain invalid <a data-cite=
-          "rfc2781#section-2">UTF-16</a> surrogates. This means the user agent
-          is free to re-encode them in any Unicode encoding (e.g.,
-          <a data-cite="rfc3629#section-3">UTF-8</a>).
+          they are not allowed to contain surrogate code points. Among other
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Will check what's going on with those Travis errors... it seems some local refs broke. Will fix them.

marcoscaceres:
There was a small regression in ReSpec that got fixed earlier today, but hasn't shipped yet:
https://github.com/w3c/respec/pull/2868

I confirmed that it's fixed by the above... should ship later today. cc @sidvishnoi.

marcoscaceres:
https://github.com/w3c/web-share/pull/154 will fix there other small ReSpec error.

PR #166, opened 2020-07-16
Introduce "web-share" feature policy

### LINKED ISSUE #151: Feature Policy integration
We should consider specifying

> The WebShare API is a [policy-controlled feature](https://w3c.github.io/webappsec-feature-policy/#policy-controlled-feature)  identified by the string "web-share". It's [default allowlist](https://w3c.github.io/webappsec-feature-policy/#default-allowlists) is 'self'. When disabled in a document, the share() method MUST return a promise which rejects with a "NotAllowedError" DOMException.

This way, the feature is disallowed by default in cross-origin documents in child browsing contexts.

ewilligers:
resolves #151

 * [X] Modified Web platform tests (https://chromium-review.googlesource.com/c/chromium/src/+/2299791)

Implementation commitment:

 * [X] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1079104)
 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1653199)

### DIFF
--- a/index.html
+++ b/index.html
@@ -394,7 +394,7 @@
     </section>
     <section id="feature-policy" data-cite="feature-policy">
       <h2>
-        Feature Policy integration
+        Permission Policy integration
       </h2>
       <p>
         This specification defines a policy-controlled feature identified by
--- a/index.html
+++ b/index.html
@@ -398,8 +398,8 @@
       </h2>
       <p>
         This specification defines a policy-controlled permission identified by
-        the string "<code><dfn>web-share</dfn></code>". Its <a>default allowlist</a> is
-        '`self`'.
+        the string "<code><dfn>web-share</dfn></code>". Its <a>default
+        allowlist</a> is '`self`'.
       </p>
       <div class="note">
         <p>
--- a/index.html
+++ b/index.html
@@ -403,7 +403,7 @@
       </p>
       <div class="note">
         <p>
-          A <a>document</a>’s feature policy determines whether a
+          A <a>document</a>’s permission policy determines whether a
           {{Navigator/share()}} call immediately rejects with a
           {{"NotAllowedError"}} {{DOMException}}.
         </p>
--- a/index.html
+++ b/index.html
@@ -398,9 +398,8 @@
       </h2>
       <p>
         This specification defines a policy-controlled permission identified by
-        the string "<code><dfn data-lt="web-share" data-nodefault=
-        "">web-share</dfn></code>". Its <a>default allowlist</a> is
-        '<code>self</code>'.
+        the string "<code><dfn>web-share</dfn></code>". Its <a>default allowlist</a> is
+        '`self`'.
       </p>
       <div class="note">
         <p>
--- a/index.html
+++ b/index.html
@@ -137,6 +137,11 @@
             |data:ShareData|, run the following steps:
           </p>
           <ol class="algorithm">
+            <li>If the current settings object's responsible document is not
+            <a>allowed to use</a> the "[=web-share-feature|web-share=]"
+            feature, return [=a promise rejected with=] with a
+            {{"NotAllowedError"}} {{DOMException}}.
+            </li>
             <li>Let |window| be [=relevant global object=] of [=this=].
             </li>
             <li>If |window| does not have [=transient activation=], return [=a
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -137,6 +137,11 @@ <h4>
             |data:ShareData|, run the following steps:
           </p>
           <ol class="algorithm">
+            <li>If the current settings object's responsible document is not
+            <a>allowed to use</a> the "[=web-share-feature|web-share=]"
+            feature, return [=a promise rejected with=] with a
```
marcoscaceres, line 142:
```suggestion
            permission, return [=a promise rejected with=] with a
```

index.html
```diff
@@ -387,6 +392,24 @@ <h3>
         </p>
       </section>
     </section>
+    <section id="feature-policy" data-cite="feature-policy">
+      <h2>
+        Feature Policy integration
```
marcoscaceres, line 397:
```suggestion
        Permission Policy integration
```

index.html
```diff
@@ -387,6 +392,24 @@ <h3>
...
+      </h2>
+      <p>
+        This specification defines a policy-controlled feature identified by
```
marcoscaceres, line 400:
```suggestion
        This specification defines a policy-controlled permission identified by
```

index.html
```diff
@@ -387,6 +392,24 @@ <h3>
...
+        the string "<code><dfn data-lt="web-share" data-nodefault=
+        "">web-share</dfn></code>". Its <a>default allowlist</a> is
+        '<code>self</code>'.
```
marcoscaceres, line 403: Not sure you need all those data-* attributes... the following should work fine.   

```suggestion
        the string "<code><dfn>web-share</dfn></code>". Its <a>default allowlist</a> is
        '`self`'.
```

index.html
```diff
@@ -387,6 +392,24 @@ <h3>
...
+      </p>
+      <div class="note">
+        <p>
+          A <a>document</a>’s feature policy determines whether a
```
marcoscaceres, line 407:
```suggestion
          A <a>document</a>’s permission policy determines whether a
```

marcoscaceres:
Gecko tracking bug https://bugzilla.mozilla.org/show_bug.cgi?id=1653199

marcoscaceres:
Testing the HTTP header on the top level doc is great, but I wonder if yo could also test it on sub-frames? See the usage of "test_feature_availability" in the following, for instance https://github.com/web-platform-tests/wpt/pull/23729/files 

It should be possible to copy/paste some of those tests.

marcoscaceres:
Blocking on test coverage.

ewilligers:
> Testing the HTTP header on the top level doc is great, but I wonder if yo could also test it on sub-frames? See the usage of "test_feature_availability" in the following, for instance https://github.com/web-platform-tests/wpt/pull/23729/files

Those tests fail in browsers. In WPTs, how do we ensure user activation in sub-frames?

Based on the linked tests, I tried adding `feature-policy/resources/feature-policy-web-share.html` .

Only when I include
`await test_driver.bless("web share");` in `feature-policy/resources/feature-policy-web-share.html`
and comment out the following lines from `resources/testdriver.js` `click`
 
```
            if (window.top !== window) {
                return Promise.reject(new Error("can only click in top-level window"));
            }
```
do I have passing tests for same-origin iframes.

marcoscaceres:
Oh good point. Let's do that as a followup. 

>  In WPTs, how do we ensure user activation in sub-frames?

... yeah, if we can't click() on subframe elements, we are kinda screwed... another way is clicking and sending a `.postMessage()`, but that only works in Chrome (it's not specified behavior in HTML yet).
 
> do I have passing tests for same-origin iframes.

Ok, that's probably good enough for now 👍

marcoscaceres:
Just FYI, I'm just adding about to send up the patch for this for Gecko (literally compiling right now 🤞). 

@othermaciej, do you want us to file WebKit bugs or is someone from WebKit tracking the changes we are making?

othermaciej:
WebKit bug(s) would be appreciated. I'd have to call on @hober for follow-through and/or soliciting a WebKit opinion on this.

marcoscaceres:
Filed https://bugs.webkit.org/show_bug.cgi?id=214448

ewilligers:
WPT issue raised for transient activation in iframes: https://github.com/web-platform-tests/wpt/issues/24755

marcoscaceres:
@ericwilligers, looks like Chrome did add some kind of use counter for this (that would count the fail cases). Could you share the results of that?  

I'd like to close #169

marcoscaceres:
Sent WebKit patch. Hopefully will land in a day or two.

PR #174, opened 2020-08-27
Reject non-http URLs for url member

### LINKED ISSUE #173: Shared URLs can be auto-fetched by native applications without respecting same-origin policy
This [public vulnerability report](https://blog.redteam.pl/2020/08/stealing-local-files-using-safari-web.html) by [Pawel Wylecial](https://www.blogger.com/profile/10114474176396848494) proposes (with proof of concept) the following attack using Web Share:

1. A malicious website shares a payload containing a `url` field whose URL would be inaccessible to the website, e.g. a `file` URL. For example: `navigator.share({url: 'file:///etc/passwd'})`.
2. The user chooses a native application, such as iOS's Mail.app, as the share target.
3. The target application automatically fetches the contents of the URL (in this example, fetching the contents of the user's password database) and attaches it to the email.
4. The user sends the email, sending the contents of their private file inadvertently, to a recipient controlled by the user.

A couple of things to note:

- This isn't specifically a vulnerability in the _browser_ (despite the above report calling it a Safari issue). The browser is just sending a well-known URL, which is effectively the same thing as sending plain text. (The URL itself is not a secret, but it points to a file whose contents is.)
- The vulnerability is due to apparent misbehaviour of the native app, which is not subject to same-origin policy and is therefore able to fetch any URL it wants, including local files. Native apps should not trust incoming intents (which could be coming from any app), but from the browser side, we have no control over the behaviour of native apps.
- This isn't very severe, as it doesn't allow the website access to the contents of the file, unless they are also able to convince the user to email it to them, by manually typing in the malicious website's email address.
- This isn't just a problem with `file` URLs. It affects any "private" URL, such as URLs located on the user's private network (e.g., router configs), and public URLs that may divulge private information if accessed with the appropriate cookies (e.g., what if the native app had cookies to a website the user stores data on, and was able to access a REST endpoint divulging user data).

Things I don't understand:

- The [blog](https://blog.redteam.pl/2020/08/stealing-local-files-using-safari-web.html) says this affects both Mail.app and Gmail on iOS. I don't understand if both Mail.app and Gmail are both doing the same behaviour and fetching the contents of the URL and setting it as an attachment, or if this is actually Safari fetching the contents of the URL and attaching it to the system intent that it sends to any application. (I am assuming it's the former, since I don't see why Safari would do the latter.)

Proposed solutions:

A. Do nothing (adding a security and privacy consideration to the spec), taking the approach that native apps should not trust incoming URLs.
B. Do nothing at the spec level, but individual implementations can filter out URLs unilaterally.
C. Add a requirement that the implementation filters out `file` URLs.
D. Add a requirement that the implementation filters out all URLs that are not same-origin as the sending site.

I think C is "theatre", it isn't going to solve the whole problem, just cover up an obvious example.

I think D is the only proper way to prevent this, but would that limit the utility of the API too severely? (It is highly likely that sites are _mostly_ using this to share same-origin URLs, but we don't have data on this.)

marcoscaceres:
Closes #173 

 * [x] Modified Web platform tests - incoming with Gecko patch.

Implementation commitment:

 * [x] WebKit (https://bugs.webkit.org/show_bug.cgi?id=215823)
 * [x] Gecko - Secure bug.

### DIFF
--- a/index.html
+++ b/index.html
@@ -35,14 +35,15 @@
           browsers: ["chrome", "firefox", "safari", "edge", "and_chr", "and_ff", "ios_saf"],
         },
         xref: "web-platform",
-        localBiblio: {
-          "Wylecial": {
-            "date": "2020-08-25",
-            "title": "Stealing local files using Safari Web Share API",
-            "Author": "Pawel Wylecial",
-            "href": "https://blog.redteam.pl/2020/08/stealing-local-files-using-safari-web.html"
-          }
-        }
+        // The Wylecial breaks the link checker with bad HTTP code response, so disabling this for now.
+        // localBiblio: {
+        //   "Wylecial": {
+        //     "date": "2020-08-25",
+        //     "title": "Stealing local files using Safari Web Share API",
+        //     "Author": "Pawel Wylecial",
+        //     "href": "https://blog.redteam.pl/2020/08/stealing-local-files-using-safari-web.html"
+        //   }
+        // }
       };
     </script>
   </head>
@@ -507,8 +508,11 @@
             when information should be confidential, so forwarding any content
             presents a risk. In particular, the {{ShareData/title}} might be
             used by an attacker to trick a user into misinterpreting the nature
-            of the content, as demonstrated in the [[Wylecial]] <a data-cite=
-            "Wylecial#">proof of concept attack</a>.
+            of the content.
+            <!--
+              , as demonstrated in the [[Wylecial]] <a data-cite=
+              "Wylecial#">proof of concept attack</a>
+            -->
           </p>
         </li>
       </ul>
--- a/index.html
+++ b/index.html
@@ -36,6 +36,14 @@
           browsers: ["chrome", "firefox", "safari", "edge", "and_chr", "and_ff", "ios_saf"],
         },
         xref: "web-platform",
+        localBiblio: {
+          "Wylecial": {
+            "date": "2020-08-25",
+            "title": "Stealing local files using Safari Web Share API",
+            "Author": "Pawel Wylecial",
+            "href": "https://blog.redteam.pl/2020/08/stealing-local-files-using-safari-web.html"
+          }
+        }
       };
     </script>
   </head>
@@ -177,8 +185,8 @@
                 <li>If |url| is failure, return <a>a promise rejected with</a>
                 {{TypeError}}.
                 </li>
-                <li>If |url|'s [=URL/scheme=] is not "http" or "https",
-                return <a>a promise rejected with</a> {{TypeError}}.
+                <li>If |url|'s [=URL/scheme=] is not "http" or "https", return
+                <a>a promise rejected with</a> {{TypeError}}.
                 </li>
                 <li>Set |data| to a copy of |data|, with its {{ShareData/url}}
                 member set to the result of running the <a>URL serializer</a>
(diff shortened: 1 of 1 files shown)


whsieh:
It looks like https://github.com/web-platform-tests/wpt/blob/master/web-share/canShare.tentative.https.html still checks that we're able to share data URLs. Would that test need to be updated as well?

ewilligers:
The webkit [patch](https://trac.webkit.org/changeset/266151/webkit) allows data urls.

@marcoscaceres   do you also consider them problematic?

martinthomson:
I think that this needs more treatment than this alone.  It needs to acknowledge the risk that share targets could retrieve content from the URL, and that they might follow redirects (or be subject to DNS tweaking) in order to exploit their position in the network to access information that might otherwise be protected by firewalls and similar.  If the application also passes on that information, that leads to an unexpected information leakage.

I don't think that it is appropriate to say that this is sufficient to address #173.

marcoscaceres:
@ericwilligers, about data: URLs, yes, as if they get dereferenced and previewed, then you can get into a situation like:

```
data:text/html;<meta http-equiv="refresh" content="1;url=file:///here/we/go/again">
``` 

@whsieh, what was the rationale for still allowing `data:` on the Webkit side? Also, you are right about `canShare()`, but that's a different matter. It's not supposed to leave the browser... but will address. I'm still hopeful we can fix `canShare()` by doing what I proposed in #108. Maybe with more urgency now? 

@martinthomson agree, definitely need to add a security considerations section for apps ingesting these URLs. As you suggested, we are just one link a weak chain here and OSs and apps able to inject need to be more careful.

marcoscaceres:
@martinthomson, I added your comment above more or less verbatim as they capture the risk quite nicely from a bunch of different angles. Please feel free to suggest additional changes.

marcoscaceres:
I file https://github.com/w3ctag/security-questionnaire/issues/96 to capture our cautionary tale of trusting URLs and thinking receiving parties will be careful with what they are receiving.

index.html
```diff
@@ -465,6 +476,21 @@ <h2>
...
+        and "https:" URL schemes. However, this mitigation is insufficient to
+        fully protect applications or operating systems that accept URLs as
+        part of a share (i.e, "share targets"). Share targets need to be
+        careful when dereferencing shared URLs for the purposes of, for
+        example, generating user-friendly previews: The risk is that share
+        targets could retrieve content from the URL, and that they might follow
+        redirects (or be subject to DNS tweaking) in order to exploit their
+        position in the network to access information that might otherwise be
+        protected by firewalls and similar. If the application also passes on
+        that information, that leads to an unexpected information leakage.
```
martinthomson, line 492: I might structure this differently for this context.  Probably with more paragraphs as well...

> Share targets that dereference a shared URL and forward that information on might forward information that might be otherwise confidential.  This can lead to unexpected information leakage if shares reference content that is only accessible by that application, the host on which it runs, or its network location.
>
> Malicious sites might exploit share targets that leak information by providing URLs that ultimately resolve to local resources, including, but not limited to, "file:" URLs or local services that might otherwise be inaccessible.  Even though this API limits shared URLS to "http:" and "https:", use of redirects to other URLs or tweaks to DNS records for hosts in those URLs might be used to cause applications to acquire content.

Here things get a little sketchy, because I'm about to make some recommendations, which I'll try to do without leaning on normative language...

> To avoid being used in these attacks, share targets can consume the URL, retrieve the content, and process that information without sharing it.  For instance, a photo editing application might retrieve an image that is "shared" with it.  A share target can also share the URL without fetching any of the referenced content.
>
> Share targets that fetch content for the purposes of offering a preview or for sharing content risk information leakage.  Content that is previewed and authorized by a user might be safe to forward, however it is not always possible for a person to identify when information should be confidential, so forwarding any content presents a risk.  In particular, the {{ShareData/title}} might be used by an attacker to trick a user into misinterpreting the nature of the content, as demonstrated in the `<a data-cite="Wylecial#">proof of concept attack</a>`.

whsieh:
> @whsieh, what was the rationale for still allowing `data:` on the Webkit side? Also, you are right about `canShare()`, but that's a different matter. It's not supposed to leave the browser... but will address. I'm still hopeful we can fix `canShare()` by doing what I proposed in #108. Maybe with more urgency now?

I believe the rationale for still allowing the `data` protocol in the change was just to keep the existing web platform tests passing.

ewilligers:
@whsieh, for the share targets you tested, is sharing a http(s) url that redirects a problem?

If not, is sharing the following simply equivalent to sharing a http(s) url that redirects,  and not equivalent to sharing a `file://` url?

```
data:text/html;<meta http-equiv="refresh" content="1;url=file:///here/we/go/again">
```

whsieh:
> @whsieh, for the share targets you tested, is sharing a http(s) url that redirects a problem?
> 
> If not, is sharing the following simply equivalent to sharing a http(s) url that redirects, and not equivalent to sharing a `file://` url?
> 
> ```
> data:text/html;<meta http-equiv="refresh" content="1;url=file:///here/we/go/again">
> ```

This is not a problem (at least, on Cocoa platforms) because neither the share targets (Messages and Mail in this case) nor the UA attempt to load the data URL as web content and then send the web content to the recipient.

index.html
marcoscaceres: I think it's quite factual and certainly doesn't read as being pushy with requirements. I made some minor tweaks.

ewilligers (approved):

PR #177, opened 2020-09-18
Add canShare() method

### LINKED ISSUE #108: Simplifying canShare()

### LINKED ISSUE #159: Explain why canShare exists at definition.
It's an unusual pattern to see .foo() and canFoo().  The specification would benefit from text or a note explaining why the canShare method exists. (As would all downstream documentation).

From a brief bit of research reading spec, MDN, and a few 'how to' articles, it isn't clear why web developers should go through the extra machinations.  Perhaps it is explained, but not near the [definition](https://w3c.github.io/web-share/level-2/#canshare-method).  There is, however, a note discussing feature detecting 'share'.

> If share() is present, there is a reasonable expectation that it will work and present the user with at least one share target. Clients can use the presence or absence of this method to determine whether to show UI that triggers its use.

Perhaps canShare is to enable an optimization to not prepare data for sharing until we know if it is possible. Likely many developers will prefer the code simplicity of feature detecting .share() and attempting the share.  This optimization doesn't seem necessary for many applications.  If there's a reason that wouldn't be a good idea, or if it specifically wouldn't work, the specification should explain why.

marcoscaceres:
Closes #108 
Closes #159 

 * [X] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/30066) 

Implementation commitment:

 * [X] WebKit - Safari 14
 * [X] Chromium - 66?  
 * [ ] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1666203)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/web-share/pull/177.html" title="Last updated on Aug 18, 2021, 7:12 AM UTC (338d7fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-share/177/1904b8f...338d7fa.html" title="Last updated on Aug 18, 2021, 7:12 AM UTC (338d7fa)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -141,6 +141,28 @@
           <h3>
             <dfn>canShare()</dfn> method
           </h3>
+          <aside class="note" title="When to use canShare()">
+            <p>
+              The {{Navigator/canShare()}} method allows developer to check if
+              a particular member is supported, and if attempting to share is
+              likely to succeed. It also works without requiring a user
+              gesture, so, unlike {{Navigator/share()}}}}, it doesn't [=consume
+              user activation=]. This is useful in deciding wether to show a
+              custom share button at all in a web application.
+            </p>
+            <pre class="example">
+              // Check if files are supported
+              const file = new File([], "empty");
+              if (navigates.canShare({files: [file]})) {
+                // enable file sharing
+              }
+
+              // Check if a URL is ok to share...
+              if (navigates.canShare({ url: someURL })) {
+                // enable basic sharing...
+              }
+            </pre>
+          </aside>
           <p>
             When the {{Navigator/canShare()}} method is called with argument
             |data:ShareData|, run the following steps:
--- a/index.html
+++ b/index.html
@@ -323,7 +323,7 @@
                 <li>If |url| is failure, return false.
                 </li>
                 <li>If |url|'s [=URL/scheme=] is not "http" or "https", return
-                return false.
+                false.
                 </li>
               </ol>
             </li>
--- a/index.html
+++ b/index.html
@@ -201,9 +201,6 @@
                 </li>
                 <li>Assert: |url| is {{URL}}.
                 </li>
-                <li>If |url|'s [=URL/scheme=] is not "http" or "https", return
-                <a>a promise rejected with</a> {{TypeError}}.
-                </li>
                 <li>Set |data| to a copy of |data|, with its {{ShareData/url}}
                 member set to the result of running the <a>URL serializer</a>
                 on |url|.
@@ -311,6 +308,9 @@
                 on |data|'s url, with |base|, and no encoding override.
                 </li>
                 <li>If |url| is failure, return false.
+                </li>
+                <li>If |url|'s [=URL/scheme=] is not "http" or "https", return
+                return false.
                 </li>
               </ol>
             </li>
--- a/index.html
+++ b/index.html
@@ -47,7 +47,7 @@
       };
     </script>
   </head>
-  <body data-cite="FILEAPI secure-contexts">
+  <body data-cite="FILEAPI secure-contexts mimesniff">
     <section id="abstract">
       <p>
         This specification defines an API for sharing text, links and other
@@ -93,18 +93,24 @@
         target to share this title and the page URL to.
       </p>
     </section>
-    <section>
+    <section data-dfn-for="Navigator">
       <h2>
         API definition
       </h2>
-      <section data-dfn-for="Navigator">
+      <section>
         <h3>
           Extensions to the `Navigator` interface
         </h3>
         <pre class="idl">
           partial interface Navigator {
-            [SecureContext] boolean canShare(optional ShareData data = {});
-            [SecureContext] Promise&lt;undefined&gt; share(optional ShareData data = {});
+            [SecureContext]
+            Promise&lt;undefined&gt; share(optional ShareData data = {});
+
+            [SecureContext]
+            boolean canShare(optional ShareData data = {});
+
+            [SecureContext]
+            boolean canShare(DOMString name, optional CanShareQuery options = {});
           };
         </pre>
         <p>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Filed [Gecko bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1666203) and implemented it. Will send patch to Mozilla.

marcoscaceres:
Working on updating the tests also.

index.html
```diff
@@ -269,6 +278,42 @@ <h4>
...
+            </li>
+            <li>If |data|'s url member is present:
+              <ol>
+                <li>Let |url:URL| be the result of running the [=URL parser]=
+                on |data|'s url, with |base|, and no encoding override.
+                </li>
+                <li>If |url| is failure, return false.
+                </li>
+                <li>If |url|'s [=URL/scheme=] is not "http" or "https", return
+                return false.
```
saschanaz, line 309:
```suggestion
                <li>If |url|'s [=URL/scheme=] is not "http" or "https", return
                false.
```

marcoscaceres:
@scheib, I added some rationale text for canShare for #159 ... would you mind taking a look and let me know what you think?

scheib:
Thank you - this addresses #159 well.  Text explaining why to use canShare looks good to me.

marcoscaceres:
@ericwilligers, @mgiuca, could we get this reviewed. It's probably the only significant thing we need to add to consider moving this spec to CR.

marcoscaceres:
Pinging again 👋 I'd like get this merge so we can move the spec to CR.

Pidut (approved):

Pidut:
Ok

Pidut:
Ok

index.html
```diff
@@ -328,6 +434,114 @@ <h3>
...
+          </dt>
+          <dd>
+            Represents support for the {{ShareData/text}} member.
+          </dd>
+        </dl>
+        <p class="Note" title="Why not an enum?">
+          The choice to use string literals instead of an [=enumeration=] is
+          deliberate. Doing so allows implementers to gracefully support future
+          {{ShareData}} member values, while also giving the user agent the
+          ability to disable sharing those members if needed.
```
saschanaz, line 542: Should this mention https://github.com/heycam/webidl/issues/893 or https://github.com/heycam/webidl/issues/491?

saschanaz:
How can we get diff preview as other w3c repositories have?

index.html
marcoscaceres: No. If something changes in WebIDL, then it should be adopted. heycam/webidl#491 should be closed anyway - it's no longer relevant.

index.html
```diff
@@ -270,6 +268,110 @@ <h4>
             or bypassing the UI if there is only a single share target.
           </div>
         </section>
+        <section>
+          <h3>
+            `canShare(data)` method
```
saschanaz, line 273: Should this be marked as historical if we think this is a broken approach?

marcoscaceres:
> How can we get diff preview as other w3c repositories have?

We might not have had .pr_preview enabled when the PR was created a long time ago... I'll see if I can get it working.

index.html
marcoscaceres: Only if we are not going to ship an alternative - but I don't think anyone plans to.

saschanaz: > Only if we are not going to ship an alternative

Are going to?

marcoscaceres: Oops, yes... there is no plans for any engine to ship anything different. Let's just ship this and be done for now. If we need a new method, we can add it when there is demand for it.

marcoscaceres:
PR preview is back 🎉

index.html
saschanaz: I don't think Gecko would want to ship anything proven to be broken as a sole solution 🤔

cc @annevk, context: https://github.com/w3c/web-share/issues/108#issuecomment-697689527

marcoscaceres: That's ok, but if we can't shift the other implementers then we need to make a process call here (i.e., Mozilla can "formally object" and we figure it out from there). 

We have two engines that already ship `.canShare()` already (and have now for over a year).

saschanaz: I think it depends on real world usage of canShare.

marcoscaceres: You are right, that too... however, we (you and I) are kinda limited in that we can't actually get that data. 

We are at an impasse here :( without input from other implementers willing to change (or even respond to issues), I don't know what we can do 😭

marcoscaceres:
Ok, [sent a PR](https://github.com/web-platform-tests/wpt/pull/30066) to remove "tentative" from the canShare() tests. The tests all pass in WebKit, except 1 that I need to look into.

marcoscaceres:
I'm going ahead and merging this as this matches two shipping implementations, and this is been sitting for nearly a year. I'm happy to continue to revise `canShare()` and even send patches to various engines if we want to improve it in various ways - so long as things remain backwards compatible. 

I'll also make another attempt to update the Gecko patch.

index.html
ewilligers: > You are right, that too... however, we (you and I) are kinda limited in that we can't actually get that data.

https://chromestatus.com/metrics/feature/timeline/popularity/2737

PR #183, opened 2020-09-24
Check transient activation after sharePromise

ewilligers:
WPT [web-share/share-sharePromise-internal-slot.https.html](https://github.com/web-platform-tests/wpt/commit/785685cf0921672462da96e99e7)
has multiple `navigator.share()` calls within a single
`onclick` handler. This is a good way to test the sharePromise
internal slot, but it was not previously permitted by the spec,
as the first call would consume activation and the later calls
would fail with `NotAllowedError` instead of `InvalidStateError`.

 * [X] Newly consistent with Web platform tests ([test](https://github.com/web-platform-tests/wpt/commit/785685cf0921672462da96e99e7))

Implementation commitment:

 * [X] Chromium (https://chromium-review.googlesource.com/c/chromium/src/+/2427944)
 * [x] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)

### DIFF
--- a/index.html
+++ b/index.html
@@ -150,6 +150,9 @@
             permission, return [=a promise rejected with=] with a
             {{"NotAllowedError"}} {{DOMException}}.
             </li>
+            <li>If {{[[sharePromise]]}} is not `null`, return <a>a promise
+            rejected with</a> {{InvalidStateError}}.
+            </li>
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):
This matches what Gecko does too, so 👍

PR #184, opened 2020-09-24
Introduce canShareFiles

ewilligers:
Note: I am not sure of the best way to word the definition.

This is an incomplete early draft seeking feedback.

Implementation commitment:

***
<a href="https://pr-preview.s3.amazonaws.com/ewilligers/web-share/pull/184.html" title="Last updated on Aug 9, 2021, 1:45 AM UTC (568ceb7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-share/184/828058d...ewilligers:568ceb7.html" title="Last updated on Aug 9, 2021, 1:45 AM UTC (568ceb7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -103,15 +103,21 @@
         </h3>
         <pre class="idl">
           partial interface Navigator {
+            [SecureContext] boolean canShareFiles(sequence&lt;USVString&gt; file_types);
             [SecureContext] Promise&lt;undefined&gt; share(optional ShareData data = {});
           };
         </pre>
         <p>
           User agents that do not support sharing SHOULD NOT expose
-          {{Navigator/share()}} on the {{Navigator}} interface.
+          {{Navigator/canShareFiles()}} or {{Navigator/share()}} on the
+          {{Navigator}} interface.
+        </p>
+        <p>
+          User agents that do not support file sharing SHOULD NOT expose
+          {{Navigator/canShareFiles()}} on the {{Navigator}} interface.
         </p>
         <div class="note">
-          The above statement is designed to permit feature detection. If
+          The above statements are designed to permit feature detection. If
           {{Navigator/share()}} is present, there is a reasonable expectation
           that it will work and present the user with at least one <a>share
           target</a>. Clients can use the presence or absence of this method to
@@ -137,6 +143,25 @@
           </dl>
         </section>
         <section>
+          <h4>
+            <dfn>canShareFiles()</dfn> method
+          </h4>
+          <p>
+            Each user agent should have an `allowlist` of permitted MIME types
+            and file extensions or file names. These may contain wildcards, for
+            example `*.txt` or `text/*`.
+          </p>
+          <p>
+            When the {{Navigator/canShareFiles()}} method is called with
+            argument |file_types|, run the following steps:
+          </p>
+          <ol class="algorithm">
+            <li>For each |file_type| in |file_types|, if |file_type| does not
+            match any entry in the user agent's `allowlist`, return `false`.
+            </li>
+            <li>Otherwise, return `true`.
+            </li>
+          </ol>
           <h4>
             <dfn>share()</dfn> method
           </h4>
@@ -194,9 +219,10 @@
                 </li>
               </ol>
             </li>
-            <li>If a file type is being blocked due to security considerations,
-            return <a>a promise rejected with</a> with a {{"NotAllowedError"}}
-            {{DOMException}}.
+            <li>If |data|'s {{ShareData/files}} member is present and
+            {{Navigator/canShareFiles()}} returns false when called with the
+            files' extensions and content types, return <a>a promise rejected
+            with</a> with a {{"NotAllowedError"}} {{DOMException}}.
             </li>
             <li>Set {{[[sharePromise]]}} to be <a>a new promise</a>.
             </li>
@@ -508,8 +534,7 @@
             when information should be confidential, so forwarding any content
             presents a risk. In particular, the {{ShareData/title}} might be
             used by an attacker to trick a user into misinterpreting the nature
-            of the content.
-            <!--
+            of the content. <!--
               , as demonstrated in the [[Wylecial]] <a data-cite=
               "Wylecial#">proof of concept attack</a>
             -->


index.html
```diff
@@ -103,15 +103,21 @@ <h3>
         </h3>
         <pre class="idl">
           partial interface Navigator {
+            [SecureContext] boolean canShareFiles(sequence&lt;USVString&gt; file_types);
```
marcoscaceres, line 106: Also, I don't see much advantage of allowing a sequence. As this is a synchronous check, it's better to only allow a single value. Then there is no ambiguity about which passed value is supported and which is not.

index.html
```diff
@@ -137,6 +143,25 @@ <h4>
           </dl>
         </section>
         <section>
+          <h4>
+            <dfn>canShareFiles()</dfn> method
+          </h4>
+          <p>
+            Each user agent should have an `allowlist` of permitted MIME types
+            and file extensions or file names. These may contain wildcards, for
+            example `*.txt` or `text/*`.
```
marcoscaceres, line 152: I think we should only support mime types, not file extensions.  

I'd be against using wildcards in the manner suggested here. Let's just stick to well-formed/valid MIME Types per:
https://mimesniff.spec.whatwg.org 

Additionally, we should have a list of allowed types that we all agree on, otherwise this API has the potential to becomes a fingerprinting vector.

index.html
```diff
@@ -194,9 +219,10 @@ <h4>
                 </li>
               </ol>
             </li>
-            <li>If a file type is being blocked due to security considerations,
-            return <a>a promise rejected with</a> with a {{"NotAllowedError"}}
-            {{DOMException}}.
+            <li>If |data|'s {{ShareData/files}} member is present and
```
marcoscaceres, line 222: The support check should be split out into its own algorithm (see #177).

index.html
```diff
@@ -137,6 +143,25 @@ <h4>
...
+            Each user agent should have an `allowlist` of permitted MIME types
+            and file extensions or file names. These may contain wildcards, for
+            example `*.txt` or `text/*`.
+          </p>
+          <p>
+            When the {{Navigator/canShareFiles()}} method is called with
+            argument |file_types|, run the following steps:
+          </p>
+          <ol class="algorithm">
+            <li>For each |file_type| in |file_types|, if |file_type| does not
```
marcoscaceres, line 159: We probably want a permissions policy check here. If not allowed, throw NotAllowedError.

index.html
```diff
@@ -194,9 +219,10 @@ <h4>
                 </li>
               </ol>
             </li>
-            <li>If a file type is being blocked due to security considerations,
-            return <a>a promise rejected with</a> with a {{"NotAllowedError"}}
-            {{DOMException}}.
+            <li>If |data|'s {{ShareData/files}} member is present and
+            {{Navigator/canShareFiles()}} returns false when called with the
+            files' extensions and content types, return <a>a promise rejected
+            with</a> with a {{"NotAllowedError"}} {{DOMException}}.
```
marcoscaceres, line 225: I think this should be a {{TypeError}} instead, as the type is not allowed to be shared.

marcoscaceres (changes requested):

marcoscaceres:
Thanks for putting this together @ericwilligers! This is a really good starting point.

index.html
```diff
@@ -137,6 +143,25 @@ <h4>
...
+            and file extensions or file names. These may contain wildcards, for
+            example `*.txt` or `text/*`.
+          </p>
+          <p>
+            When the {{Navigator/canShareFiles()}} method is called with
+            argument |file_types|, run the following steps:
+          </p>
+          <ol class="algorithm">
+            <li>For each |file_type| in |file_types|, if |file_type| does not
+            match any entry in the user agent's `allowlist`, return `false`.
```
marcoscaceres, line 160: For the mime types, they need to be [parsed](https://mimesniff.spec.whatwg.org/#parse-a-mime-type) (which can return "failure", so throw a TypeError), and then the "[essence](https://mimesniff.spec.whatwg.org/#mime-type-essence)" of each type can be checked against the allowlist.

index.html
marcoscaceres: Maybe we should generalize the name too... to `canShareType()`

marcoscaceres, line 106: We probably just want a regular `DOMString`. 

```suggestion
            [SecureContext] boolean canShareFiles(sequence&lt;DOMString&gt; file_types);
```

index.html
marcoscaceres: Note here, we need to do the same check as in #177 to make sure `{files: []}` gets rejected, but passing `{text: "hi", files: []}` is ok.

index.html
```diff
@@ -194,9 +219,10 @@ <h4>
                 </li>
               </ol>
             </li>
-            <li>If a file type is being blocked due to security considerations,
-            return <a>a promise rejected with</a> with a {{"NotAllowedError"}}
-            {{DOMException}}.
+            <li>If |data|'s {{ShareData/files}} member is present and
+            {{Navigator/canShareFiles()}} returns false when called with the
+            files' extensions and content types, return <a>a promise rejected
```
marcoscaceres, line 224: Note: This needs to be spec'ed as a `for each |file:File| in files:`, the check each file's content type separately.

marcoscaceres:
Prototype implementation in Gecko using `.canShareType()`: 

Screen Shot 2020-09-28 at 1 36 52 pm

I like it :)

index.html
annevk: I agree this needs more details, also for web developers who should have some assurances about the capabilities.

NotWoods:
We discussed `canShareFiles` and TAG concerns about adding many methods to the Navigator namespace. If future share categories arise in the future (i.e.: formatted text, html) we would want a `canShareX` for each which is not ideal.

A design we settled on is to change `canShare` to take two parameters, a category and optional extra data to refine the check as the second parameter.

```idl
partial interface Navigator {
  boolean canShare(ShareCategory category, optional any option);
};

enum ShareCategory {
  "files"
}
```

The type of the second parameter would change based on the first argument, similar to `eventTarget.addEventListener`. In the case of `"files"`, it would correspond to the `file_type` parameter in this proposal. 

```js
// check if file sharing is supported
if (navigator.canShare("files") { ... }

// check if image sharing is supported
if (navigator.canShare("files", "image/*") { ... }
```

marcoscaceres:
Re: ` optional any option);`, I'd prefer we use a dictionary there. Then we can have the `types:` specifically defined there: `sequence<DOMString> types;`.

marcoscaceres:
Actually, playing around a bit... just `DOMString type;` might be more useful than `sequence types;`.... otherwise, we get back into the situation where you need to test arrays with a single item to work out what is actually supported.

marcoscaceres:
I've updated https://github.com/w3c/web-share/pull/177 to specify `canShare(name, options)`... for backwards compat, what we might be able to do is support both:

```WebIDL
            [SecureContext]
            boolean canShare(optional ShareData data = {});

            [SecureContext]
            boolean canShare(DOMString name, optional CanShareQuery options = {});    
```

Then we can just encourage developers to use the `canShare(name, options)` version when it's widely supported. 

Let me know what you think...

NotWoods:
Can we simply leave the previous version as a Chromium-only variant rather than including it in the spec?

marcoscaceres:
> Can we simply leave the previous version as a Chromium-only variant rather than including it in the spec?

We can, but that might end up having web compat issues (for Gecko)... I guess it somewhat depends on what the Webkit folks want to do also. Safari shipped with `canShare()` matching Chrome, I believe. 

The above allows us to move away from the original version in somewhat sane way without breaking too much content (although the usage stats suggest that usage of `.canShare()` is low still). 

cc @othermaciej

marcoscaceres:
Closing, as we not implemented by anyone. Looks like we are sticking with `.canShare()`.

Issue #178, opened 2020-09-22
Limiting to http/https is limiting


martinthomson:
#173 was a bit of a learning experience.  The fix for that in #174 limits URLs to http or https.  That was the most conservative reaction to that issue.

This means that a number of URI schemes with no known exposure to this vulnerability can no longer be shared.  From standard protocols mailto, sip[s], and tel are all examples of schemes that do not invoke actions and should be safe to share.  (Why you might use those rather than share a vCard resource is not relevant; these seem valid to me.)  Then there are the quasi-standard things like: ipfs, magnet.  And a whole bunch of proprietary schemes: acrobat, zoommtg, steam, microsoft-edge (ok, maybe we don't want that...), and so forth.

Now that the dust has settled, it might be worth examining principles a little closer to determine whether a looser set of constraints can be made to work.  @dveditz suggested that we block URLs if we might not permit both navigation or redirection.  I think that is a reasonable starting point here.  We don't allow navigation or redirects to `file://` and so sharing that seems to be primarily a means of circumventing that policy.

We can't expect that share targets will respect the same policies that a browser does in case it follows an HTTP URL that redirects to file:///, but we are explicitly accepting that risk already by allowing use of HTTP schemes.

marcoscaceres:
The above sounds good. I'd like to get #177 (`canShare()` method) landed first and get all UAs passing the associated tests. Then we should be in a good position to reconsider what we did in #173.

tomayac:
Maybe we need to mint a new [absolute](https://url.spec.whatwg.org/#absolute-url-string) and [relative URL string](https://url.spec.whatwg.org/#relative-url-string) that *excludes* `file:`, but includes everything else what's currently listed for the [allowed value of `url`](https://w3c.github.io/web-share/#dom-sharedata-url).

martinthomson:
Excluding `file:` is not sufficient.  The patch I'm in the process of landing (@marcoscaceres, details on request) blocks blob, javascript, chrome, about, wss, and data.  These are things that Firefox knows about and would not permit a top-level navigation to, as I originally stated.  If we find more of these and teach Firefox not to navigate to them, those would be blocked by virtue of that.

[...]

Issue #127, opened 2019-09-27
Explicitly state that UA can reject file based on its type


mgiuca:
With examples to prevent sharing of executable files.

Could be based on extension, MIME type or contents.

mgiuca:
It actually does have this (search for the word "block").

But it needs to change to split it out from the user activation check, and move it into the "in parallel" steps, so it can be done async.

marcoscaceres:
Over on the gecko side, we've been considering the following attack:

```JS
const evil = new File([...evilBytes], "cat.exe");
navigator.share({files: [evil], title: "Cute cat image!"});
```

In order to mitigate the above, we should require each `File` have both type and a filename with an extension, and that the type must match an expected extension. We should have a limited list of recommended sharable types/extension combinations. 

This also allows us to avoid having to do content type sniffing, although that could be done as well. 

So, concretely, we should replace:

> If a file type is being blocked due to security considerations, return a promise rejected with with a "NotAllowedError" DOMException.

With a lookup-table for allowed MIME + extensions-list combinations.

mgiuca:
TPAC 2020: Resolved to make a change to the spec to let browsers determine their own rules for blocking based on file type.

We can discuss further having a formal list of allowed / blocked types, but no consensus on that.

@NotWoods suggests https://html.spec.whatwg.org/multipage/links.html#allowed-to-download as a starting point.

marcoscaceres:
The HTML suggestion relates to the `download=` attribute, but in our case we use Permissions Policy to enable the functionality. However, I adapted the text "Optionally, the user agent may return false, if it believes doing so would safeguard the user from a potentially hostile download." and added it to https://github.com/w3c/web-share/pull/185/commits/6c0a9f5a90a4a35375f3211c385be380c73b5a62.

marcoscaceres:
List of extensions / types from Chromium : https://docs.google.com/document/d/1tKPkHA5nnJtmh2TgqWmGSREUzXgMUFDL6yMdVZHqUsg/edit

marcoscaceres:
closing as handled by https://github.com/w3c/web-share/commit/6c0a9f5a90a4a35375f3211c385be380c73b5a62 for now.

PR #185, opened 2020-09-28
Add "validate share data" algorithm

marcoscaceres:
This keeps the good parts we derived from "canShare()". 

 * [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/30046). The other parts are already tested. 
 
Implementation commitment:

 * [x] [WebKit](https://sourcegraph.com/github.com/WebKit/WebKit/-/blob/Source/WebCore/page/Navigator.cpp?L127)
 * [x] Chromium

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/web-share/pull/185.html" title="Last updated on Aug 17, 2021, 5:51 AM UTC (892497f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-share/185/e35a1d2...892497f.html" title="Last updated on Aug 17, 2021, 5:51 AM UTC (892497f)">Diff</a>


index.html
```diff
@@ -269,6 +258,55 @@ <h4>
...
+            present.
+            </li>
+            <li>If |data|'s {{ShareData/files}} member is present:
+              <ol>
+                <li>If |titleTextOrUrl| is false, and if |data|'s
+                {{ShareData/files}} member is empty, or return false.
+                  <p class="note">
+                    This causes a `{ files: [] }` dictionary to be treated as
+                    an empty dictionary. However, passing a dictionary like
+                    `{text: "text" files: []}` is fine, as `files` is just
```
martinthomson, line 285:
```suggestion
                    `{text: "text", files: []}` is fine, as `files` is just
```

index.html
```diff
@@ -269,6 +258,55 @@ <h4>
...
+            present, return false.
+            </li>
+            <li>Let |titleTextOrUrl:boolean| be true if any of
+            {{ShareData/title}}, or {{ShareData/text}}, or {{ShareData/url}} is
+            present.
+            </li>
+            <li>If |data|'s {{ShareData/files}} member is present:
+              <ol>
+                <li>If |titleTextOrUrl| is false, and if |data|'s
+                {{ShareData/files}} member is empty, or return false.
```
martinthomson, line 281:
```suggestion
                {{ShareData/files}} member is empty, return false.
```

index.html
```diff
@@ -269,6 +258,55 @@ <h4>
...
+            {{ShareData/text}}, or {{ShareData/url}} or {{ShareData/files}} are
+            present, return false.
+            </li>
+            <li>Let |titleTextOrUrl:boolean| be true if any of
+            {{ShareData/title}}, or {{ShareData/text}}, or {{ShareData/url}} is
+            present.
+            </li>
+            <li>If |data|'s {{ShareData/files}} member is present:
+              <ol>
+                <li>If |titleTextOrUrl| is false, and if |data|'s
```
martinthomson, line 280:
```suggestion
                <li>If |titleTextOrUrl| is false and |data|'s
```

index.html
```diff
@@ -269,6 +258,55 @@ <h4>
...
+                  <p class="note">
+                    This causes a `{ files: [] }` dictionary to be treated as
+                    an empty dictionary. However, passing a dictionary like
+                    `{text: "text" files: []}` is fine, as `files` is just
+                    ignored.
+                  </p>
+                </li>
+                <li>If the implementation does not support file sharing, return
+                false.
+                </li>
```
martinthomson, line 291: Is this the point where you want to add validation for file content type/extension?

marcoscaceres: yes, this would be the place.... open question:

 1. Should we require both the filename and the type? Or should be require just one or the other. I'm leaning towards only requiring the filename, which must include an "allowed extension". If the type is included, it must be one that maps to the type (i.e., ".jpg" maps to "image/jpg" or "image/jpeg"). 
 1. Should we allow dot files (e.g,., ".png")? My feeling is no because of how they behave in some OSs.

martinthomson: An issue discussion is probably a better place to discuss that.  Should we use an issue?

marcoscaceres:
@ericwilligers, @mgiuca, I think this is still useful to merge as a basis on which to build the things we discussed during the call. It's mostly just a refactor based on @inexorabletash suggestions.

index.html
marcoscaceres: Yes, let's pick this up in https://github.com/w3c/web-share/issues/127

index.html
```diff
@@ -270,6 +263,60 @@ <h4>
...
+            present, return false.
+            </li>
+            <li>Let |titleTextOrUrl:boolean| be true if any of
+            {{ShareData/title}}, or {{ShareData/text}}, or {{ShareData/url}} is
+            present.
+            </li>
+            <li>If |data|'s {{ShareData/files}} member is present:
+              <ol>
+                <li>If |titleTextOrUrl| is false and |data|'s
+                {{ShareData/files}} member is empty, return false.
```
saschanaz, line 286: This line is new, right? Maybe better to split it out as a new PR for easier tracking?

index.html
```diff
@@ -195,7 +188,7 @@ <h4>
               </ol>
             </li>
             <li>If a file type is being blocked due to security considerations,
-            return <a>a promise rejected with</a> with a {{"NotAllowedError"}}
+            return [=a promise rejected with=] with a {{"NotAllowedError"}}
```
saschanaz, line 191:
```suggestion
            return [=a promise rejected with=] a {{"NotAllowedError"}}
```

index.html
```diff
@@ -463,6 +510,10 @@ <h2>
         is only [=exposed=] in [=secure contexts=] (such as `https://`
         schemes).
         </li>
+        <li>Third-party contexts require that they be [=allowed to use=] the
+        {{Navigator/share()}} method is one of the mechanisms described in
+        [[[permissions-policy]]].
```
saschanaz, line 515: This also looks new, maybe better to split out.

index.html
marcoscaceres: I think this should stay. The spec already deals with files, so it's appropriate to check this here.

index.html
marcoscaceres: https://github.com/w3c/web-share/pull/210

index.html
saschanaz: I mean, `If |titleTextOrUrl| is false` part is new and thus does changes behavior.

marcoscaceres: Ah, no, this is just a refactor of step 6 and 7 of the current spec:

6 .If none of data's members title, text, or url or files are present, return a promise rejected with a TypeError.
7. If data's files member is present:
7.1 If data's files member is empty, or if the implementation does not support file sharing, return a promise rejected with a TypeError, and abort these steps.

saschanaz: The current spec says `{ text: "foo", files: [] }` is invalid, and this patch makes it valid by checking other field values.

marcoscaceres: Ah right, still. Respectfully, I'd like to keep this as is.

marcoscaceres:
Ok, my intent it to merge this tomorrow as it matches Chrome/WebKit's behavior. 

We can then continue with #177 and figuring out #178.

saschanaz:
Okay, thanks for https://github.com/web-platform-tests/wpt/pull/30046.

saschanaz (approved):

marcoscaceres:
> Okay, thanks for web-platform-tests/wpt#30046.

No problem. Do you want to review those too by chance?

PR #220, opened 2021-09-20
fix: canShare() return false if not allowed to use

marcoscaceres:

 * [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/30868)

Implementation commitment:

 * [x] WebKit (https://bugs.webkit.org/show_bug.cgi?id=214448)
 * [ ] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=1251175)
 * [x] Gecko (https://phabricator.services.mozilla.com/D90834)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/web-share/pull/220.html" title="Last updated on Sep 20, 2021, 10:35 PM UTC (9a32b1d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-share/220/94e7883...9a32b1d.html" title="Last updated on Sep 20, 2021, 10:35 PM UTC (9a32b1d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -307,6 +307,9 @@
             </li>
             <li>If |document| is not [=Document/fully active=], return false.
             </li>
+            <li>If |document| is not <a>allowed to use</a> <a>"web-share"</a>,
+            return `false`.
+            </li>
             <li>Return the result of [=validate share data=] with |data| and
             [=this=]'s [=relevant settings object=]'s [=environment settings
             object/API base URL=].
--- a/index.html
+++ b/index.html
@@ -308,7 +308,7 @@
             <li>If |document| is not [=Document/fully active=], return false.
             </li>
             <li>If |document| is not <a>allowed to use</a> <a>"web-share"</a>,
-            return `false`.
+            return false.
             </li>
             <li>Return the result of [=validate share data=] with |data| and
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -307,6 +307,9 @@ <h3>
             </li>
             <li>If |document| is not [=Document/fully active=], return false.
             </li>
+            <li>If |document| is not <a>allowed to use</a> <a>"web-share"</a>,
+            return `false`.
```
marcoscaceres, line 311:
```suggestion
            return false.
```

saschanaz:
Do we have WPT for this BTW?

marcoscaceres:
I'll add one. No problem.

marcoscaceres:
Added/fixed test:
https://github.com/web-platform-tests/wpt/pull/30868
 
> BTW I'm not sure what the enum version of canShare be look like, if canShare covers more than just data check 🤔

Happy to build on that aspect incrementally as a followup. 

Irrespective, I think it would be confusing for ".canShare(x)" return `true` and then calling `.share()` to fail. That would not be great. This keeps them in sync.

Issue #222, opened 2021-11-15
Share not display image


DEVfancybear:
After sharing, only the link is displayed, not the uploaded image.

marcoscaceres:
Hi @DEVfancybear, this repository is for specification, not for a particular browser. However, if you tell us which browser you are having an issue in, we might be able to point you in the right direction.

DEVfancybear:
@marcoscaceres I'm getting an error sharing images but only showing the link in safari browser, can you help me

marcoscaceres:
If you put the share() code up somewhere, I can check if it’s a browser bug or not.

DEVfancybear:
![code](https://user-images.githubusercontent.com/47390684/141873649-920e0d4f-7e89-4428-b57e-6b92176d942c.png)
I have tested this code on adnroid and it works as I want, but when using on ios safari, when I share the image, I only get the link, for example telegram or skype are not displayed.

marcoscaceres:
Thanks @DEVfancybear. I can't see anything obviously wrong, so it might be a bug or limitation in Safari. 

I'll see if I can track down someone from the WebKit team to see if they can reproduce the issue.

DEVfancybear:
![9](https://user-images.githubusercontent.com/47390684/142366969-33bab7fe-53cc-465c-80e2-542b1968b30b.png)
![8](https://user-images.githubusercontent.com/47390684/142366991-fb4cc2ee-08a6-4895-8f4e-25fe97b0c437.png)
If I select only update vs file, not url, text, title, will the new iphone device like iphone 12 promax show such error @marcoscaceres ?

DEVfancybear:
![9](https://user-images.githubusercontent.com/47390684/142367242-0058dbf4-68c9-4431-b762-cc9b787d8088.png)
If I select only update vs file, not url, text, title, will the new iphone device like iphone 12 promax show such error @marcoscaceres ?

marcoscaceres:
@DEVfancybear could you please put the actual file you are testing again somewhere? I.e., it may be that the browser supports it, but it's not actually shareable for other reasons.

DEVfancybear:
But in the picture is the error Unsupported feature: navigator.canShare() @marcoscaceres ?

marcoscaceres:
@DEVfancybear please understand that it’s not possible for me to replicate the problem without all the same code+data.

saschanaz:
The demo looks wrong btw. It shouldn't fail with canShare-is-unsupported message.

[...]

PR #226, opened 2021-11-19
Better log messages for canShare

saschanaz:
Related to #222

### DIFF
--- a/demos/share-files.html
+++ b/demos/share-files.html
@@ -118,10 +118,20 @@
         textfield.value = '';
     }
 
+    function checkBasicFileShare() {
+      // XXX: There is no straightforward API to do this.
+      // For now, assume that text/plain is supported everywhere.
+      const txt = new Blob(['Hello, world!'], {type: 'text/plain'});
+      // XXX: Blob support? https://github.com/w3c/web-share/issues/181
+      const file = new File([txt], "test.txt");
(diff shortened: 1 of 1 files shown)


demos/share-files.html
```diff
@@ -118,10 +118,20 @@ <h1>
         textfield.value = '';
     }
 
+    function checkBasicFileShare() {
+      // XXX: There is no straightforward API to do this.
```
saschanaz, line 122: I think this is a problem, as already mentioned in https://github.com/w3c/web-share/issues/108#issuecomment-697689527. Something like this would help.

```js
navigator.canShare("files"); // basic feature support
navigator.canShare("files", [files]); // matches canShare({files});
```

saschanaz: https://github.com/w3c/web-share/pull/184 was about `canShareFiles` with file types, but since Blink is now also checking file sizes, I guess receiving a sequence of files is needed?

saschanaz: canShare in Blink seemingly has no check for file types nor file sizes, and share() somehow throws PermissionError instead. Maybe to prevent fingerprinting? But then what's the point of canShare with the current form? 👀

marcoscaceres: hehe, "incremental improvement"?

marcoscaceres (approved):

PR #236, opened 2022-05-27
Editorial: use global task to resolve promises

### LINKED ISSUE #235: Promises cannot be manipulated while "in parallel"
You need to use "queue a global task" and from that task you can manipulate promises. (This applies to state in general, but it seems this API mainly uses promises.)

marcoscaceres:
Closes #235

Editorial fix-up of what we originally intended. Doesn't affect UA behavior is and shouldn't be observable/testable (AFAIK).

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 27, 2022, 7:07 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fweb-share%2Fc1e3ac98a686aa46f3fd9fdb59ba9c6ed871da8a%2Findex.html%3FisPreview%3Dtrue)

```
📡 HTTP Error 404: http://labs.w3.org/spec-generator/uploads/flncu8
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/web-share%23236.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -188,12 +188,12 @@
             return [=a promise rejected with=] an {{"InvalidStateError"}}
             {{DOMException}}.
             </li>
-            <li>Let |window| be [=relevant global object=] of [=this=].
+            <li>Let |global:Window| be [=this=]'s [=relevant global object=].
             </li>
-            <li>If |window| does not have [=transient activation=], return [=a
+            <li>If |global| does not have [=transient activation=], return [=a
             promise rejected with=] a {{"NotAllowedError"}} {{DOMException}}.
             </li>
-            <li>[=Consume user activation=] of |window|.
+            <li>[=Consume user activation=] of |global|.
             </li>
             <li>Let |base:URL| be [=this=]'s <a>relevant settings object</a>'s
             [=environment settings object/API base URL=].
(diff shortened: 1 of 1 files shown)


ewilligers (approved):

marcoscaceres:
Thanks for reviewing it @ewilligers 🙏

marcoscaceres:
Taking as is... @annevk, please let us know if we missed anything.

annevk:
This looks good. State changes after reject/resolve are okay here as when script runs the state changes will have been made, though if you ever do something else in these tasks you want to be careful that does not change. (E.g., if you start dispatching events you need to make the state changes first.)

marcoscaceres:
Appreciate the guidance, @annevk! will keep an eye our for this also in other specs (already spotted the same issue a few times in Payment Request!).

PR #240, opened 2022-06-07
Editorial: move UI suggestion to Priv/Sec section

### LINKED ISSUE #64: Move UI note to Sec/Priv section
The note that states:

> always shows some form of UI, to give the user a choice of application...

Would probably better to move that to the Security and privacy considerations. Also, the language there seems to overstep a bit into making hard UI requirements. 

I suggest softening the language a bit, and make recommendations instead.

marcoscaceres:
closes #64

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/web-share/pull/240.html" title="Last updated on Jul 7, 2022, 5:15 AM UTC (93ed578)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-share/240/3f97331...93ed578.html" title="Last updated on Jul 7, 2022, 5:15 AM UTC (93ed578)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -309,8 +309,9 @@
                 </li>
                 <li>Present the user with a choice of one or more <a>share
                 targets</a>, selected at the user agent's discretion. The user
-                MUST be given the option to cancel rather than choosing any of
-                the share targets. Wait for the user's choice.
+                agent MUST give the user the option to cancel rather than
+                choosing any of the [=share targets=]. Wait for the user's
+                choice.
                 </li>
                 <li>If the user chose to cancel the share operation, [=queue a
                 global task=] on the [=user interaction task source=] using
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -625,11 +614,15 @@ <h2>
...
-        <li>There is a requirement to not allow the website to learn which apps
-        are installed, or which app was chosen from {{Navigator/share()}},
-        because this information could be used for fingerprinting, as well as
+        <li>The API does not expose to the website which [=share targets=] are
+        available, or which share target was chosen by the user from
+        {{Navigator/share()}}. Doing so could be used for fingerprinting by
         leaking details about the user's device.
         </li>
+        <li>User agents are discouraged from showing any kind of "always use
+        this target in the future" option, or bypassing the UI if there is only
```
ewilligers, line 623: The wording here regarding "bypassing the UI" being discouraged is not strong enough.

"prohibited" might have been unreasonably strong, but I think it better captured the intent.

marcoscaceres: I wonder if we should flip this saying that it should always require user interaction irrespective of how many share targets are available.  

Note that we already do this normatively as part of `.share()`; this section just emphasizes what we already normatively require.

marcoscaceres: Just to clarify, this section is just guidance. We shouldn't/can't really "prohibit" anything. 

Maybe just "strongly discourage"?

ewilligers: "should always require user interaction" is fine. That makes the intent clear.

marcoscaceres: We already normatively require "transient activation" activation to use web share (and it's verifiable for conformance). I. can state that again though as a statement of fact.

ewilligers (approved):

PR #244, opened 2022-07-01
Define sharable scheme + check

### LINKED ISSUE #178: Limiting to http/https is limiting

marcoscaceres:
Closes #178

Implementation commitment:

### DIFF
--- a/index.html
+++ b/index.html
@@ -397,6 +397,21 @@
             Validate share data
           </h3>
           <p>
+            A <dfn>sharable scheme</dfn> is any of the following [=URL=]
+            [=URL/schemes=]:
+          </p>
+          <ol>
+            <li>`http`
+            </li>
+            <li>`https`
+            </li>
+            <li>`data`
+            </li>
+            <li>Any [=safelisted scheme=] that the user agent supports for the
+            purpose of sharing.
+            </li>
+          </ol>
+          <p>
             To <dfn>validate share data</dfn> with |data:ShareData| and
             |base:URL|, run the following steps:
           </p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -438,13 +451,11 @@ <h3>
...
                 <li>If |url| is failure, return false.
                 </li>
-                <li>
-                  <p>
-                    If |url| is a URL the user agent deems potentially hostile
-                    (e.g., "file:") or wouldn't make sense to outside the scope
-                    of the document (e.g., "blob:"), return false.
-                  </p>
-                  <aside class="issue" data-number="178"></aside>
+                <li>If the |url|'s [=URL/scheme=] is a [=local scheme=], or
```
marcoscaceres, line 454: I know this check is kinda redundant, because they are not in the the "sharable scheme" list, but I wanted to be explicit in banning them.

index.html
```diff
@@ -438,13 +451,11 @@ <h3>
...
+                `file`, or `javascript`, or `wss`, return false.
```
marcoscaceres, line 455:
```suggestion
                `file`, or `javascript`, or `ws`, or `wss`, return false.
```

ewilligers (approved):

ewilligers (approved):

marcoscaceres:
Actually, turned out we have tests already. https://github.com/web-platform-tests/wpt/blob/master/web-share/share-url-invalid.https.html

